PR #28, opened 2024-05-15
Entity types

adithyadinesh0412:


src/.env.sample
```diff
@@ -118,3 +118,18 @@ ALLOWED_HOST = "http://exampleDomain.com"
 
 #Authentication meathod to be used
 AUTH_METHOD=jwt_only #or user_service_authenticated
```
ankit1shahu: Auth method we can say only 'jwt' no need to say jwt_only
@adithyadinesh0412

src/controllers/v1/config.js
```diff
@@ -0,0 +1,13 @@
+const configsService = require('@services/config')
+
+module.exports = class configs {
```
ankit1shahu, line 3: can you add some description for better understanding what it will do
@adithyadinesh0412

src/database/models/organizationExtensions.js
```diff
@@ -0,0 +1,49 @@
...
+	const organizationExtensions = sequelize.define(
+		'organizationExtensions',
+		{
+			id: {
+				type: DataTypes.INTEGER,
+				allowNull: false,
+				primaryKey: true,
+				autoIncrement: true,
+			},
+			organization_id: {
```
ankit1shahu: it should be unique @adithyadinesh0412

src/database/models/organizationExtensions.js
```diff
@@ -0,0 +1,49 @@
...
+				type: DataTypes.INTEGER,
+				allowNull: false,
+			},
+			review_required: {
```
ankit1shahu: why the default value is false if show_reviewer_list is on that means you will be showing reviewer but by making false user can directly publish resources on the platform. If org admin wants to make any changes then he can update but by default we should keep it true.
@adithyadinesh0412 please confirm with @Prajwal17Tunerlabs as well

src/database/models/organizationExtensions.js
```diff
@@ -0,0 +1,49 @@
...
+			min_approval: {
+				type: DataTypes.INTEGER,
+				allowNull: false,
+			},
+			resource_type: {
+				type: DataTypes.INTEGER,
+				allowNull: false,
+			},
+			review_type: {
+				type: DataTypes.STRING,
```
ankit1shahu: we can have enum value will be selected from sequential and parallel only please check that @adithyadinesh0412

src/database/queries/entityType.js
```diff
@@ -0,0 +1,136 @@
...
+				where: filter,
+				...options,
+				raw: true,
+			})
+		} catch (error) {
+			throw error
+		}
+	}
+
+	static async findAllEntityTypes(orgIds, attributes, filter = {}) {
```
ankit1shahu: this function can not be called without org id and will return only org entity we can rename this function to findOrgEntityTypes @adithyadinesh0412

ankit1shahu: we can pass the org id in the filter only why do we have to pass specifically @adithyadinesh0412 
Then this function can utilized other place as well

src/database/queries/entityType.js
```diff
@@ -0,0 +1,136 @@
...
+	static async findUserEntityTypesAndEntities(filter) {
+		try {
+			const entityTypes = await EntityType.findAll({
+				where: filter,
+				raw: true,
+			})
+
+			const entityTypeIds = entityTypes.map((entityType) => entityType.id)
+
+			const entities = await Entity.findAll({
```
ankit1shahu: instead of importing entity model here we can create function in entity query file we can call from here @adithyadinesh0412

src/database/queries/entityType.js
```diff
@@ -0,0 +1,136 @@
...
+				const matchingEntities = entities.filter((entity) => entity.entity_type_id === entityType.id)
+				return {
+					...entityType,
+					entities: matchingEntities,
+				}
+			})
+
+			return result
+		} catch (error) {
+			console.error('Error fetching data:', error)
```
ankit1shahu: we have logger we can use that for logging please don't use console log @adithyadinesh0412

src/database/queries/entityType.js
```diff
@@ -0,0 +1,136 @@
...
+	static async findEntityTypeById(filter) {
+		try {
+			return await EntityType.findByPk(filter)
+		} catch (error) {
+			return error
+		}
+	}
+
+	static async findAllEntityTypesAndEntities(filter) {
+		try {
```
ankit1shahu: whats difference between above function findUserEntityTypesAndEntities and this function we can use only one function and we can pass different filters @adithyadinesh0412

src/envVariables.js
```diff
@@ -155,6 +155,30 @@ let enviromentVariables = {
 		optional: true,
 		default: common.AUTH_METHOD.JWT_ONLY,
 	},
+	REVIEW_REQUIRED: {
```
ankit1shahu: In model you have added default value false and here true it should be sync if user is not passing then @adithyadinesh0412

src/services/config.js
```diff
@@ -0,0 +1,95 @@
...
+	 * List Configs.
+	 * @method
+	 * @name list
+	 * @param {Integer} organization_id
+	 * @param {String} id -  id.
+	 * @returns {JSON} - List of configs as response.
+	 */
+
+	static async list(organization_id) {
+		let orgExtenstionData = {}
```
ankit1shahu: why variables are declared outside the try block @adithyadinesh0412

src/services/entity-types.js
```diff
@@ -0,0 +1,239 @@
...
+	 * Create entity type.
+	 * @method
+	 * @name create
+	 * @param {Object} bodyData - entity type body data.
+	 * @param {String} id -  id.
+	 * @returns {JSON} - Created entity type response.
+	 */
+
+	static async create(bodyData, id, orgId) {
+		bodyData.created_by = id
```
ankit1shahu: why is this outside try block @adithyadinesh0412

src/services/config.js
```diff
@@ -0,0 +1,95 @@
...
+		let configData = []
+		try {
+			// define filter
+			const filter = {
+				organization_id,
```
ankit1shahu: @adithyadinesh0412 please optimize this function

src/services/entity-types.js
```diff
@@ -0,0 +1,239 @@
...
+	 * @method
+	 * @name update
+	 * @param {Object} bodyData -  body data.
+	 * @param {String} id - entity type id.
+	 * @param {String} loggedInUserId - logged in user id.
+	 * @returns {JSON} - Updated Entity Type.
+	 */
+
+	static async update(bodyData, id, loggedInUserId, orgId) {
+		bodyData.updated_by = loggedInUserId
```
ankit1shahu: please declare all the variable in try block only @adithyadinesh0412

ankit1shahu (changes requested):
Reviewed On 17 May 4:30

src/.env.sample
adithyadinesh0412: Its the same in mentoring, may be we can discuss this and take it up in a different PR

src/database/models/organizationExtensions.js
adithyadinesh0412: Removing this page as it's already added by priyanka

src/database/queries/entityType.js
adithyadinesh0412: So as you said in the previous comment, let the orgId be mandatory for this function and in the filter we can pass any additional filters we have. So to fetch the entity_types , org_id is a mandatory field.

adithyadinesh0412: Removing this function as you said in the next comment. we can reuse findAllEntityTypesAndEntities()

src/constants/endpoints.js
```diff
@@ -1 +1,3 @@
-module.exports = {}
+module.exports = {
+	ORGANIZATION_READ: 'v1/organization/read',
```
priyanka-TL: @adithyadinesh0412 this endpoint may vary in different user services, add this in the env

src/controllers/v1/entity-types.js
```diff
@@ -0,0 +1,80 @@
+// Dependencies
+const entityTypeService = require('@services/entity-types')
+
+module.exports = class Entity {
```
priyanka-TL, line 4: @adithyadinesh0412 change it to EntityType

src/controllers/v1/entity-types.js
```diff
@@ -0,0 +1,80 @@
...
+	 * @name update
+	 * @param {Object} req - request data.
+	 * @returns {JSON} - entities updating response.
+	 */
+
+	async update(req) {
+		try {
+			return await entityTypeService.update(
+				req.body,
+				req.params.id,
```
priyanka-TL, line 32: @adithyadinesh0412 params.id will have 1st priority

src/constants/endpoints.js
adithyadinesh0412: Sure

src/.env.sample
```diff
@@ -129,4 +129,7 @@ SHOW_REVIEWER_LIST =true
 MIN_APPROVAL=1
 
 # instance level configuration regarding review type can be  SEQUENTIAL / PARALLEL
-REVIEW_TYPE = "sequential"
\ No newline at end of file
+REVIEW_TYPE = "sequential"
+
+# mention the organization read API to the curresponding service 
+ORGANIZATION_READ = 'v1/organization/read'
```
priyanka-TL: @adithyadinesh0412 change it to ORGANIZATION_READ_ENDPOINT

src/services/entity-types.js
```diff
@@ -0,0 +1,239 @@
...
+	/**
+	 * Create entity type.
+	 * @method
+	 * @name create
+	 * @param {Object} bodyData - entity type body data.
+	 * @param {String} id -  id.
+	 * @returns {JSON} - Created entity type response.
+	 */
+
+	static async create(bodyData, id, orgId) {
```
priyanka-TL: @adithyadinesh0412 change this id to loggedInUserId

PR #44, opened 2024-05-29
module apis added

priyanka-TL:


src/controllers/modules.js
```diff
@@ -0,0 +1,73 @@
+const modulesService = require('@services/modules')
+
```
adithyadinesh0412: This file should come under controllers/v1 right? @priyanka-TL

adithyadinesh0412:
Approved 4th June

PR #49, opened 2024-06-06
Comment ap is

priyanka-TL:


src/constants/common.js
```diff
@@ -66,4 +66,6 @@ module.exports = {
 	SORT_ORDER: 'sort_order',
 	URL: 'url',
 	FILE_PATH: 'filePath',
+	STATUS_RESOLVED: 'RESOLVED',
+	ALL_USER_ROLES: 'all',
```
adithyadinesh0412, line 70: We have FILTER_ALL key also with the same value. Can we reuse / combine both? @priyanka-TL

priyanka-TL: no let be separate

adithyadinesh0412: ok

PR #71, opened 2024-06-27
Permissions check added in authenticator

adithyadinesh0412:


src/constants/common.js
```diff
@@ -86,4 +86,5 @@ module.exports = {
 	STATUS_RESOLVED: 'RESOLVED',
 	ALL_USER_ROLES: 'all',
 	LIMIT: 100,
+	roleValidationPaths: [],
```
priyanka-TL: @adithyadinesh0412 why roleValidationPaths path is empty?

PR #104, opened 2024-07-22
added req changes and project create changes

priyanka-TL:


adithyadinesh0412:
Approved.

src/services/projects.js
```diff
@@ -529,4 +552,82 @@ module.exports = class ProjectsHelper {
...
+				common.PROJECT,
+				loggedInUserId
+			)
+
+			const url = getSignedUrl.result[resourceId].files[0].url
+			const blobPath = getSignedUrl.result[resourceId].files[0].file
+
+			let config = {
+				method: 'put',
+				maxBodyLength: Infinity,
```
aks30: Let's put an env configurable max size.

src/services/projects.js
```diff
@@ -529,4 +552,82 @@ module.exports = class ProjectsHelper {
...
+
+	/**
+	 * Upload to cloud
+	 * @method
+	 * @name uploadToCloud
+	 * @param {Object} req.id - project id
+	 * @returns {JSON} - project delete response.
+	 */
+
+	static async uploadToCloud(resourceId, loggedInUserId, orgId, bodyData) {
```
aks30: @priyanka-TL Will this override the file every time? Can you add the bucket path of the file as a comment in code.

src/services/projects.js
```diff
@@ -529,4 +552,82 @@ module.exports = class ProjectsHelper {
...
 	}
+
+	/**
+	 * Upload to cloud
+	 * @method
+	 * @name uploadToCloud
+	 * @param {Object} req.id - project id
+	 * @returns {JSON} - project delete response.
+	 */
+
```
aks30: @priyanka-TL this function should just upload the file - tomorrow same can be used by program/observation, setting the path etc. It should not update the project table here,it can be done in the calling function.

src/services/projects.js
```diff
@@ -283,7 +245,68 @@ module.exports = class ProjectsHelper {
 					response.result &&
 					Object.keys(response.result).length > 0
 				) {
-					result = { ...result, ...response.result }
+					//modify the response as label value pair
+					let resultData = response.result
+					const userDetails = {
+						organization_id: orgId,
+						id: loggedInUserId,
```
aks30: @priyanka-TL why do we need userid to get entity types and entities?

aks30 (changes requested):
Reviewed on 22 July 5:30 PM

src/services/projects.js
priyanka-TL: yes

src/.env.sample
```diff
@@ -143,5 +143,9 @@ USER_PROFILE_DETAILS_ENDPOINT='v1/user/read'
 #Release 1.0
 #available resource types
 RESOURCE_TYPES="projects,observations,observation_with_rubrics,surveys,programs"
+
 #maximum number of tasks allowed in a project
 MAX_PROJECT_TASK_COUNT= 10
+
+#maximum body length of file
+MAX_BODY_LENGTH_FOR_UPLOAD=Infinity
```
aks30: @priyanka-TL We should give other sample values as well and in what scale it will be - Byte, KB, MB?

src/envVariables.js
```diff
@@ -142,6 +142,11 @@ let enviromentVariables = {
 		optional: false,
 		default: 10,
 	},
+	MAX_BODY_LENGTH_FOR_UPLOAD: {
+		message: 'Maximum body length for file upload is required',
+		optional: false,
+		default: Infinity,
```
aks30: @priyanka-TL - Let's put some logical value of few MBs or something.

src/services/projects.js
```diff
@@ -49,13 +49,54 @@ module.exports = class ProjectsHelper {
...
 				const mappingData = {
 					resource_id: projectCreate.id,
 					creator_id: loggedInUserId,
 					organization_id: orgId,
 				}
 				await resourceCreatorMappingQueries.create(mappingData)
+
+				//upload to blob
+				const resourceId = projectCreate.id
+				const fileName = `${loggedInUserId}${resourceId}${orgId}project.json`
```
aks30: @priyanka-TL why do we need an org id here? can we change it to just userId/resourceId.json?
 Also are we using private bucket or public? Which package do we use for cloud upload? We should be using a common CSP package being used in Elevate project and a pvt bucket only. If not we should take that up as a separate task.

src/services/projects.js
```diff
@@ -125,53 +166,50 @@ module.exports = class ProjectsHelper {
...
 					id: resourceId,
 					organization_id: orgId,
 				}
+
 				let updateData = {
-					meta: { title: bodyData.title },
-					title: bodyData.title,
 					updated_by: loggedInUserId,
-					blob_path: getSignedUrl.result[resourceId].files[0].file,
+					blob_path: projectUploadStatus.blob_path,
```
aks30, line 190: @priyanka-TL Why do we have to update blob path, it will be the same once created, no?

src/services/projects.js
```diff
@@ -283,7 +321,65 @@ module.exports = class ProjectsHelper {
...
+									const match = validEntities.find((entity) => entity.value === value.toLowerCase())
+									resultData[key] = match || { label: value, value: value.toLowerCase() }
+								}
+							}
+						}
+
+						result = { ...result, ...resultData }
+					}
 				}
 			}
```
aks30, line 391: @priyanka-TL Do they need an org name in the details API?

aks30 (changes requested):
Reviewed on 23 July 8:30 AM

src/services/projects.js
```diff
@@ -529,4 +625,46 @@ module.exports = class ProjectsHelper {
...
+				data: JSON.stringify(bodyData),
+			}
+
+			let projectUploadStatus = await axios.request(config)
+			return {
+				blob_path: blobPath,
+				result: projectUploadStatus,
+			}
+		} catch (error) {
+			console.log(error, 'error')
```
aks30: @priyanka-TL Remove console log

src/services/projects.js
```diff
@@ -529,4 +625,46 @@ module.exports = class ProjectsHelper {
...
+
+	/**
+	 * Upload to cloud
+	 * @method
+	 * @name uploadToCloud
+	 * @param {Object} req.id - project id
+	 * @returns {JSON} - upload project response.
+	 */
+
+	static async uploadToCloud(fileName, resourceId, loggedInUserId, bodyData) {
```
aks30: @priyanka-TL Move this to a common module will be used for other resources as well, this function can be a generic function and has no specific mention of project

src/services/projects.js
priyanka-TL: Not required

src/services/projects.js
priyanka-TL: While creating the project we are not updating blob path because the resource id is required to generate file name and path. hence after creating we are updating the db

src/services/projects.js
priyanka-TL: removed orgId. the package we are using is **client-cloud-services**. we are using private bucket.

PR #116, opened 2024-07-30
File validation

priyanka-TL:


src/.env.sample
```diff
@@ -155,3 +155,6 @@ RESOURCE_AUTO_SAVE_TIMER="30000"
 
 # max length for the notes to the reviewer 
 MAX_RESOURCE_NOTE_LENGTH="256"
+
+# max number of the evidences allowed
+MAX_NO_OF_EVIDENCE_ALLOWED=10
```
aks30: @priyanka-TL what is the key used for?

src/envVariables.js
```diff
@@ -157,6 +157,11 @@ let enviromentVariables = {
 		optional: false,
 		default: 256, //256 characters
 	},
+	MAX_NO_OF_EVIDENCE_ALLOWED: {
+		message: 'Max number of evidences allowed is required',
+		optional: false,
```
aks30: If default value is given we should make it optional true

src/services/projects.js
```diff
@@ -470,131 +523,334 @@ module.exports = class ProjectsHelper {
...
+				['id', 'value', 'validations', 'has_entities']
+			)
+
+			const taskEntityTypesMapping = taskEntityTypes.reduce((acc, item) => {
+				acc[item.value] = item
+				return acc
+			}, {})
+
+			//get all entity type validation for file types
+			let allowedFileTypesEntityTypes = await entityTypeQueries.findUserEntityTypeAndEntities(
```
aks30: Why is the same query done here and on line 527, file validations also should have models

src/services/projects.js
```diff
@@ -470,131 +523,334 @@ module.exports = class ProjectsHelper {
...
 						throw {
-							message: `${entityType.value} not added`,
-							error: utils.errorObject(common.BODY, entityType.value),
+							error: utils.errorObject(common.BODY, entityType.value, `${entityType.value} is required`),
+						}
+					}
+				}
+
+				// check max character limit exceeded
+				if (Object.keys(entityType.validations).includes(common.MAX_CHARACTER_LIMIT) && fieldData) {
```
aks30: @priyanka-TL Why do we need to do like this, it can all be done via regex - https://www.sitepoint.com/using-regular-expressions-to-check-string-length/

aks30 (changes requested):
Reviewed on 31 July

src/services/projects.js
priyanka-TL: i will change it

src/services/projects.js
priyanka-TL: Already regex is there for special character check, so we cant send valid messages, hence added one more key in the validation itself.

src/.env.sample
priyanka-TL: for setting the maximum number of evidences which is allowed in the resource

aks30: As discussed it should be via regex.

src/database/migrations/20240514075609-add-default-entities.js
```diff
@@ -8,6 +8,13 @@ module.exports = {
 		}
 
 		const entitiesArray = [
+			{
+				entityType: 'title',
+				entities: '',
+				has_entities: false,
+				validation: { regex: `[^A-Za-z0-9 <>_&-]`, required: true },
```
aks30, line 15: @priyanka-TL The max allow length should be checked via regex here. For all entity types length has to be checked via regex including NOTES

src/envVariables.js
```diff
@@ -157,6 +157,11 @@ let enviromentVariables = {
 		optional: false,
 		default: 256, //256 characters
 	},
+	MAX_NO_OF_EVIDENCE_ALLOWED: {
```
aks30: @priyanka-TL -As discussed we can remove this.

src/package.json
```diff
@@ -38,6 +38,7 @@
 		"fn-args": "^4.0.0",
 		"form-data": "^4.0.0",
 		"fs-extra": "^9.0.1",
+		"i18n": "^0.15.0",
```
aks30, line 41: @priyanka-TL check if this is needed.

src/services/projects.js
```diff
@@ -99,11 +100,19 @@ module.exports = class ProjectsHelper {
 					throw new Error('FILE_UPLOADED_FAILED')
 				}
 			} catch (error) {
-				return responses.failureResponse({
-					message: error.message || error,
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
-				})
+				if (error.name === 'SequelizeDatabaseError' && error.original.code === '22001') {
+					return responses.failureResponse({
```
aks30: @priyanka-TL we should let them know the key for which this validation is coming.

src/services/projects.js
```diff
@@ -215,11 +224,19 @@ module.exports = class ProjectsHelper {
 				throw new Error('FILE_UPLOADED_FAILED')
 			}
 		} catch (error) {
-			return responses.failureResponse({
-				message: error.message || error,
-				statusCode: httpStatusCode.bad_request,
-				responseCode: 'CLIENT_ERROR',
-			})
+			if (error.name === 'SequelizeDatabaseError' && error.original.code === '22001') {
+				return responses.failureResponse({
```
aks30: @priyanka-TL  we should let them know the key for which this validation is coming.

src/services/projects.js
aks30: @priyanka-TL As discussed you will add it in regex and not via env

src/services/projects.js
```diff
@@ -470,131 +523,319 @@ module.exports = class ProjectsHelper {
...
-					)
-					task.learning_resources.forEach((learningResource) => {
-						let validateURL = utils.checkRegexPattern(subTaskEntityTypes[0], learningResource.url)
-						if (validateURL) {
-							throw {
-								message: 'INCORRECT_LEARNING_RESOURCE',
-								error: utils.errorObject(common.BODY, common.CHILDREN),
+				//check child task validations
+				if (task.children && task.children.length > 0) {
+					for (let childTask of task.children) {
```
aks30, line 598: @priyanka-TL As discussed we can have this as a common function, only change is the name of model and if subtask model we should have parent_id as mandatory.

aks30 (changes requested):
Reviewed on 1 Aug 11:50 AM

src/database/migrations/20240514075609-add-default-entities.js
priyanka-TL: Note limit need to send in config api hence keeping this in env

src/generics/utils.js
```diff
@@ -477,12 +492,23 @@ const convertToInteger = (value) => {
 	return isNaN(value) ? false : parseInt(value, 10)
 }
 
+const compareLength = (value1, value2) => {
```
aks30: @priyanka-TL do we need this still ?

src/generics/utils.js
```diff
@@ -477,12 +492,23 @@ const convertToInteger = (value) => {
...
+	return returnValue
+}
+
 const isLabelValuePair = (item) => {
 	return Array.isArray(item)
 		? item.every((subItem) => subItem && typeof subItem === 'object' && 'label' in subItem && 'value' in subItem)
 		: typeof item === 'object' && 'label' in item && 'value' in item
 }
 
+const validateTitle = (title) => {
```
aks30, line 502: @priyanka-TL Let me know if this is still needed

src/services/projects.js
```diff
@@ -470,131 +527,129 @@ module.exports = class ProjectsHelper {
...
-					let checkRegex = utils.checkRegexPattern(entityType, fieldData)
-					if (checkRegex) {
-						throw {
-							message: `Special characters not allowed in ${entityType.value}`,
-							error: utils.errorObject(common.BODY, entityType.value),
-						}
+			//validate project data
+			for (let entityType of entityTypes) {
+				let validationResult = await this.validateEntityData(
+					projectData,
```
aks30: @priyanka-TL just pass the required entityType here, also can we do this in parallel?

src/services/projects.js
```diff
@@ -470,131 +527,129 @@ module.exports = class ProjectsHelper {
...
-
-			entityTypes.forEach((entityType) => {
-				const fieldData = projectData[entityType.value]
+			//fetch task entityType validations
+			let taskEntityTypes = await entityModelMappingQuery.findEntityTypesAndEntities(
+				{
+					model: common.TASKS,
+					status: common.STATUS_ACTIVE,
+				},
+				userDetails.organization_id,
```
aks30: @priyanka-TL I think we should use should user resource org id , what if a user is move to an another org but he can work. on the resource he has created in previous org right?

src/services/projects.js
```diff
@@ -470,131 +527,129 @@ module.exports = class ProjectsHelper {
...
 
-					if (taskEntityType.has_entities) {
-						let checkEntities = utils.checkEntities(taskEntityType, fieldData)
-						if (!checkEntities.status) {
-							throw {
-								message: checkEntities.message,
-								error: utils.errorObject(common.TASKS, taskEntityType.value),
-							}
+			// validate task
+			for (let task of projectData.tasks) {
```
aks30, line 580: @priyanka-TL we can do this in parallel

src/services/projects.js
```diff
@@ -604,19 +659,142 @@ module.exports = class ProjectsHelper {
...
+					}
+				} else {
+					let checkRegex = utils.checkRegexPattern(entityType, fieldData)
+					if (checkRegex) {
+						return {
+							hasError: true,
+							error: utils.errorObject(
+								sourceType,
+								entityType.value,
+								`${model} ${entityType.value} is invalid, please ensure it contains no special characters and does not exceed the character limit`
```
aks30, line 781: @priyanka-TL Let's put a validationMessage column in entity type table and use that, I am very sure people would want to change this later. Field name can be added via here.

aks30 (changes requested):
Reviewed on 1 Aug 8 PM

src/generics/utils.js
priyanka-TL: yes required for project creation

src/services/projects.js
priyanka-TL: the same entity type is used by different model

PR #136, opened 2024-08-13
Start Review API

priyanka-TL:


src/services/reviews.js
```diff
@@ -120,11 +121,14 @@ module.exports = class reviewsHelper {
 				await handleComments(bodyData.comment, resourceId, userId)
 			}
 
+			// update the status if review assigned then update status as STARTED else CHANGES_REQUESTED
+			let status = common.REVIEW_STATUS_REQUESTED_FOR_CHANGES
+			if (startReview) {
+				status = common.REVIEW_STATUS_STARTED
+			}
```
adithyadinesh0412: Can we do this with ternary operator?  @priyanka-TL 

const status = startReview ? common.REVIEW_STATUS_STARTED : common.REVIEW_STATUS_REQUESTED_FOR_CHANGES

adithyadinesh0412: Also consider using const instead of let if its value is not updated elsewhere.

adithyadinesh0412 (changes requested):
Reviewed on 13th Aug

src/services/reviews.js
adithyadinesh0412: consider using const if the value is not updated elsewhere.

adithyadinesh0412 (approved):
Approved on 13 Aug

src/controllers/v1/reviews.js
```diff
@@ -7,22 +7,43 @@
...
 	 * @param {Object} req - review body data.
-	 * @returns {JSON} - Review creation or update details
+	 * @returns {JSON} - Review update response
 	 */
 
 	async update(req) {
 		try {
 			const updateReview = await reviewService.update(
-				req.params.id, //resource id
+				req.params.id,
```
adithyadinesh0412, line 43: Please add validators for these new apis @priyanka-TL

src/services/projects.js
```diff
@@ -634,22 +635,22 @@ module.exports = class ProjectsHelper {
...
 				//written as a backup will remove once the user service PR merged
 				if (Array.isArray(reviewers?.data?.result?.data) && reviewers.data.result.data.length > 0) {
 					reviewerIds = reviewers.data.result.data.map((item) => item.id)
+				} else {
+					// If no valid reviewers data is found, return an error response
+					throw new Error('REVIEWER_IDS_NOT_FOUND')
 				}
 
+				//return error message if the reviewer is invalid or not found
+				if (bodyData.reviewer_ids.length >= reviewers.data.result.data.length)
```
adithyadinesh0412: This have an edge case,  
since bodyData can be passed from the frontend , they can repeat the reviewer_ids. but while searching we get only unique values. So in such a case we cannot go for length match. 
Ex : 
in body if we receive 
[1,2,2,3,4,5,5] length will be 7
and the result from user service will be 5.

Try this or any other feasible method 
const areArraysEqual = _.isEqual(_.sortBy(array1), _.sortBy(array2));

@priyanka-TL

src/services/resource.js
```diff
@@ -914,7 +914,7 @@ module.exports = class resourceHelper {
 
 		let resourceWiseLevels = {}
 
-		if (reviewLevelDetails) {
+		if (reviewLevelDetails.length > 0) {
```
adithyadinesh0412, line 917: reviewLevelDetails is an array or object? @priyanka-TL

src/services/reviews.js
```diff
@@ -98,49 +158,43 @@ module.exports = class reviewsHelper {
...
-				return responses.failureResponse({
-					message: 'REVIEW_NOT_FOUND',
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
-				})
-			}
+			if (!review?.id) throw new Error('REVIEW_NOT_FOUND')
+
+			//check the review is already started
+			if (review.status === common.RESOURCE_STATUS_STARTED) throw new Error('REVIEW_ALREADY_STARTED')
```
adithyadinesh0412: STARTED is a review status, can we update that? @priyanka-TL

adithyadinesh0412 (changes requested):
Reviewed on 14th Aug

src/services/resource.js
priyanka-TL: @adithyadinesh0412  findAll will return an array findOne will return object

adithyadinesh0412: Ok got it.

src/services/reviews.js
```diff
@@ -98,49 +158,43 @@ module.exports = class reviewsHelper {
...
-				return responses.failureResponse({
-					message: 'REVIEW_NOT_FOUND',
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
-				})
-			}
+			if (!review?.id) throw new Error('REVIEW_NOT_FOUND')
+
+			//check the review is already started
+			if (review.status === common.REVIEW_STATUS_STARTED) throw new Error('REVIEW_ALREADY_STARTED')
```
adithyadinesh0412, line 166: Can u also check if there are any other dependancy with this key? in the existing code?

adithyadinesh0412: @priyanka-TL

adithyadinesh0412 (approved):
Approved - 14 th Aug

PR #138, opened 2024-08-15
Activity tracker

priyanka-TL:


src/controllers/v1/actions.js
```diff
@@ -0,0 +1,32 @@
...
+	 * @name update
+	 * @param {Integer} id  action id.
+	 * @param {Object} body  action data
+	 * @returns {JSON} - Detail of action as response.
+	 */
+	async update(req) {
+		try {
+			let action
+			if (req.params.id) {
+				action = await actionService.update(req.params.id ? req.params.id : '', req.body)
```
adithyadinesh0412: We have `if (req.params.id)` check. so again a ternary operator in the function call is redundant. @priyanka-TL

src/database/migrations/20240815181652-add-permission-for-actions.js
```diff
@@ -0,0 +1,78 @@
...
+					module: 'actions',
+					request_type: ['POST', 'GET'],
+					api_path: '/scp/v1/actions/*',
+					created_at: new Date(),
+					updated_at: new Date(),
+					created_by: 0,
+				},
+				{
+					role_title: common.ORG_ADMIN_ROLE,
+					permission_id: await getPermissionId('actions', ['POST', 'GET'], '/scp/v1/actions/*'),
```
adithyadinesh0412: Should the ORG_ADMIN have access to actions apis? 
Since it has no org based flows and common for all organization should we give this access to the org_admins?
@priyanka-TL

priyanka-TL: its same like modules org admin also can add their own actions

src/validators/v1/actions.js
```diff
@@ -0,0 +1,45 @@
...
+		req.checkParams('id')
+			.trim()
+			.optional()
+			.notEmpty()
+			.withMessage('id param is empty')
+			.isNumeric()
+			.withMessage('id param is invalid, must be an integer')
+
+		req.checkBody('code')
+			.notEmpty()
```
adithyadinesh0412, line 26: Since we are passing id , can we make code optional?

priyanka-TL: Whatever database field is mandatory to create record and it doesn't have default values, we can add that as a mandatory param in req body. So it wont throw error while creation or updation

src/database/migrations/20240813123008-create-actions-table.js
```diff
@@ -0,0 +1,43 @@
+'use strict'
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+	async up(queryInterface, Sequelize) {
+		await queryInterface.createTable('actions', {
```
adithyadinesh0412, line 6: Can we add paranoid to do a soft delete in this table? @priyanka-TL

src/database/migrations/20240814125518-add-default-actions.js
```diff
@@ -0,0 +1,67 @@
...
+				updated_at: timestamp,
+			}
+		})
+
+		// Insert the data into the 'actions' table
+		await queryInterface.bulkInsert('actions', actionsFinalArray)
+	},
+
+	down: async (queryInterface, Sequelize) => {
+		await queryInterface.bulkDelete('actions', null, {})
```
adithyadinesh0412, line 65: Can we delete only specific items we added while reverting the migration?
In case we are reverting the data added after a point where we have actual user created data , we might loose it. 
@priyanka-TL

src/database/migrations/20240815181652-add-permission-for-actions.js
```diff
@@ -0,0 +1,110 @@
...
+					permission_id: await getPermissionId('actions', ['POST', 'GET', 'DELETE'], '/scp/v1/actions/*'),
+					module: 'actions',
+					request_type: ['POST', 'GET', 'DELETE'],
+					api_path: '/scp/v1/actions/*',
+					created_at: new Date(),
+					updated_at: new Date(),
+					created_by: 0,
+				},
+				{
+					role_title: common.ORG_ADMIN_ROLE,
```
adithyadinesh0412, line 81: Should we give delete / update permission to ORG_ADMIN? Since the activities are not org centric and common for the entire system. @priyanka-TL

src/database/models/activities.js
```diff
@@ -0,0 +1,48 @@
...
+			},
+		},
+		{
+			modelName: 'Activity',
+			tableName: 'activities',
+			freezeTableName: true,
+			paranoid: true,
+			indexes: [
+				{
+					fields: ['organization_id', 'user_id'],
```
adithyadinesh0412: Should we index action id also? In case if we have to search based on any specific action like created / deleted ? @priyanka-TL

src/helpers/requester.js
```diff
@@ -0,0 +1,52 @@
...
+
+const post = async (baseURL, route = '', headers = {}, body = {}, queryParams = {}) => {
+	try {
+		baseURL = baseURL.replace(/\/$/, '')
+		let url = baseURL + route
+		if (!utils.isEmpty(queryParams)) url += '?' + new URLSearchParams(queryParams).toString()
+		console.log('URL: ', url)
+		console.log('HEADERS: ', JSON.stringify(headers, null, 2))
+		console.log('BODY: ', JSON.stringify(body, null, 2))
+		const response = await axios.post(url, body, { headers, timeout: 3000 })
```
adithyadinesh0412: can we have the timeout value from common? @priyanka-TL

src/database/migrations/20240814125518-add-default-actions.js
priyanka-TL: @adithyadinesh0412 As it a first migration, we are adding the first data hence we can bulk delete all. If we are adding one action then in the down action we can delete only that time,

src/database/migrations/20240815181652-add-permission-for-actions.js
priyanka-TL: @adithyadinesh0412  That is not required, its not org centric, but still org admin can create new and enhance or delete one item and instead of that they can add new item. Its there choice, only thing is if they are adding new actions they need to write the code for that

src/helpers/requester.js
adithyadinesh0412, line 1: We have a file generic/request.js which handles post and get requests. How is this different from that? @priyanka-TL

src/database/models/activities.js
priyanka-TL: We cannot delete the activity of any user. Only thing is we can list it, For list as of now org and user id is enough

priyanka-TL: Its like a tracker, we cant delete anything

src/helpers/requester.js
priyanka-TL: It is specific to events

src/database/migrations/20240815181652-add-permission-for-actions.js
adithyadinesh0412: @aks30  Having a confusion would like to have your thoughts here.
So here we are giving the permission to ORG_ADMINs to do CRUD operations on Actions table. 
Since actions are not org centric I'm doubting if we have to do that or not. 
My point is; what if an org admin of `A` changes the action description from one language to other or any text change specific to his/her org. It will be reflected throughout the actions in the system which is not right in a SAAS model. 
So should we allow the org admins to do that or should we bring in org concept in actions?

adithyadinesh0412 (approved):

src/controllers/v1/activities.js
```diff
@@ -0,0 +1,39 @@
...
+
+	async list(req) {
+		try {
+			let userId = req.decodedToken.id
+			let orgId = req.decodedToken.organization_id
+			if (utils.validateRoleAccess(req.decodedToken.roles, common.ADMIN_ROLE)) {
+				userId = req.query.user_id ? req.query.user_id : req.decodedToken.id
+				orgId = req.query.organization_id ? req.query.organization_id : req.decodedToken.id
+			} else if (utils.validateRoleAccess(req.decodedToken.roles, common.ORG_ADMIN_ROLE)) {
+				userId = req.query.user_id ? req.query.user_id : req.decodedToken.id
```
aks30: @priyanka-TL why do we need this condition? Any role who has this permission should be able to access this API

src/controllers/v1/projects.js
```diff
@@ -23,7 +22,11 @@ module.exports = class Projects {
 			if (req.params.id) {
 				let project = {}
 				if (req.method === common.REQUEST_METHOD_DELETE) {
-					project = await projectService.delete(req.params.id, req.decodedToken.id)
+					project = await projectService.delete(
+						req.params.id,
+						req.decodedToken.id,
+						req.decodedToken.organization_id
```
aks30: @priyanka-TL Why is this added for activity tracker PR?

aks30: Because if this is for activity tracker we should take org id from project/resource

src/database/migrations/20240813123014-create-activities-table.js
```diff
@@ -0,0 +1,55 @@
...
+			updated_at: {
+				allowNull: false,
+				type: Sequelize.DATE,
+			},
+			deleted_at: {
+				type: Sequelize.DATE,
+			},
+		})
+		//add index
+		await queryInterface.addIndex('activities', ['organization_id', 'user_id'], {
```
aks30: @priyanka-TL in which file do we define creating of distribution columns?

src/database/models/activities.js
```diff
@@ -0,0 +1,48 @@
...
+			},
+			user_id: {
+				allowNull: false,
+				type: DataTypes.STRING,
+			},
+			action_id: {
+				allowNull: false,
+				type: DataTypes.INTEGER,
+			},
+			object_id: {
```
aks30, line 19: @priyanka-TL should object id be indexed, because for a given resource people might want all the activities

src/validators/v1/actions.js
```diff
@@ -0,0 +1,45 @@
...
+	update: (req) => {
+		req.checkParams('id')
+			.trim()
+			.optional()
+			.notEmpty()
+			.withMessage('id param is empty')
+			.isNumeric()
+			.withMessage('id param is invalid, must be an integer')
+
+		req.checkBody('code')
```
aks30: @priyanka-TL - I agree with @adithyadinesh0412 only what is required or needs to be updated should be passed in update call, in this code might not be required.

src/services/actions.js
```diff
@@ -0,0 +1,174 @@
...
+	 * Create Action
+	 * @method
+	 * @name create
+	 * @param {String} bodyData - action creation data
+	 * @returns {JSON} - action creation response
+	 */
+	static async create(bodyData) {
+		try {
+			const action = await actionQueries.findOne({ code: bodyData.code })
+			if (action?.id) {
```
aks30: @priyanka-TL why do we have to do this check? It will automatically fail because of the DB index right?

src/controllers/v1/comments.js
```diff
@@ -46,7 +46,8 @@ module.exports = class comments {
 				req.params.id ? req.params.id : '',
 				req.query.resource_id,
 				req.body,
-				req.decodedToken.id
+				req.decodedToken.id,
+				req.decodedToken.organization_id
```
aks30: @priyanka-TL do we not need to pick from resource the org id?

src/helpers/eventBroadcaster.js
```diff
@@ -0,0 +1,33 @@
...
+		queryParams = {},
+	}
+) => {
+	try {
+		const endPoints = events.eventEndpoints[action]
+		await Promise.all(
+			endPoints.map((endPoint) => {
+				if (endPoint.type === common.EVENT_TYPE_API) {
+					if (endPoint.method === 'POST')
+						requester.post(endPoint.baseUrl, endPoint.route, headers, requestBody, queryParams)
```
aks30: @priyanka-TL are we making API calls to to broad cast events? It should rather be internal event broad casting.

E.g.- https://nodejs.org/api/events.html, https://medium.com/deno-the-complete-reference/10-use-cases-of-event-emitters-in-node-js-d88c0b9bb536

src/services/comments.js
```diff
@@ -22,10 +25,28 @@ module.exports = class CommentsHelper {
...
 	 * @param {Object} bodyData - Request Body
 	 * @param {String} userId - User ID
+	 * @param {String} orgId - Organization ID
 	 * @returns {JSON} - comment id
 	 */
-	static async update(commentId = '', resourceId, bodyData, userId) {
+	static async update(commentId = '', resourceId, bodyData, userId, orgId) {
 		try {
+			//validate resource
+			const resource = await resourceQueries.findOne(
```
aks30, line 32: @priyanka-TL how is this related to activity tracker?

src/services/projects.js
```diff
@@ -253,57 +264,65 @@ module.exports = class ProjectsHelper {
...
-				return responses.failureResponse({
-					message: 'PROJECT_NOT_FOUND',
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
-				})
+				throw new Error('PROJECT_NOT_FOUND')
 			}
+
+			//add user action
+			eventBroadcaster(common.EVENT_ADD_USER_ACTION, {
```
aks30: @priyanka-TL Rather then adding like this in multiple places, can we write a hook in model? E.g. - https://github.com/ELEVATE-Project/mentoring/blob/master/src/database/models/modules.js

aks30 (changes requested):
Reviewed on 24 Aug

src/validators/v1/actions.js
priyanka-TL: @aks30 update API is for create and update action. I made the changes, if id is there code and description i made as optional when id is there

src/controllers/v1/comments.js
priyanka-TL: it is passed for create activity, but its not required now because the review status will be inprogress when click on start button

src/helpers/eventBroadcaster.js
priyanka-TL: @aks30  this is same like mentoring, currently they are using events for internal api calls, now i added a function call also into that, so both api or function call will work, when call the publish api we can use this event

src/services/projects.js
priyanka-TL: No we cant add this, hence update API is an auto save call, hence activities will add in each call.

src/controllers/v1/activities.js
priyanka-TL: @aks30 i added because currently i gave this permission to admin and org admin, 
org admin can see any user activity, admin can see any org any user activity.

i will remove it and add permission to all scp users

src/services/comments.js
priyanka-TL: Not a part of activity, It was part of validation PR by mistake committed to here, to validate the status of resource

src/services/projects.js
aks30: @priyanka-TL we can do for comments model and for resource model we can do the same by checking status, if status not equal to draft or something

src/helpers/eventBroadcaster.js
aks30: @priyanka-TL I don't think the mentoring service uses events, user service might and they do API calls because they have to update the mentoring service if a user or org is updated. We don't need all that here, keep it to what we need only.

src/database/models/resources.js
```diff
@@ -92,6 +94,38 @@ module.exports = (sequelize, DataTypes) => {
...
+					orgId: instance.organization_id,
+				})
+			}
+		} catch (error) {
+			console.error(`Error during ${actionType} hook:`, error)
+			throw error
+		}
+	}
+
+	Resource.addHook('afterCreate', (instance) => emitUserAction(instance, 'RESOURCE_CREATED'))
```
adithyadinesh0412, line 115: Can we have RESOURCE_CREATED and RESOURCE_DELETED from common? @priyanka-TL

src/services/actions.js
```diff
@@ -20,11 +20,6 @@ module.exports = class ActionsHelper {
 	 */
 	static async create(bodyData) {
 		try {
-			const action = await actionQueries.findOne({ code: bodyData.code })
```
adithyadinesh0412: Can we check if the is already created? @priyanka-TL

src/database/models/resources.js
adithyadinesh0412: As discussed resolving this

adithyadinesh0412 (approved):

src/controllers/v1/activities.js
```diff
@@ -0,0 +1,34 @@
...
+	 * Get user activities
+	 * @method
+	 * @name list
+	 * @param {String} req.pageNo - Page No.
+	 * @param {String} req.pageSize - Page size limit.
+	 * @param {String} req.searchText - Search text.
+	 * @returns {JSON} - activities List.
+	 */
+
+	async list(req) {
```
aks30, line 22: @priyanka-TL To include resource id filter

src/database/models/actions.js
```diff
@@ -0,0 +1,52 @@
...
+			tableName: 'actions',
+			freezeTableName: true,
+			paranoid: true,
+		}
+	)
+
+	Action.addHook('beforeDestroy', async (instance, options) => {
+		try {
+			// Soft-delete associated Activities records with matching actions
+			await sequelize.models.Activity.update(
```
aks30: @priyanka-TL We will not delete any activity if user action is deleted. In action delete API check if there is any activity for that action, if yes then don't allow to delete that action.

src/services/reviews.js
```diff
@@ -260,7 +259,7 @@ module.exports = class reviewsHelper {
 
 			// Publish resource if isPublishResource is true
 			if (isPublishResource) {
-				const publishResource = await this.publishResource(resourceId, resource.user_id)
+				const publishResource = await this.publishResource(resourceId, resource.user_id, orgId)
```
aks30: @priyanka-TL to remove here and everywhere else where org is not required and can be derived from the resource in the hook of the model?

src/distributionColumns.psql
```diff
@@ -9,4 +9,5 @@ SELECT create_distributed_table('resources', 'organization_id');
 SELECT create_distributed_table('resource_creator_mapping', 'creator_id');
 SELECT create_distributed_table('review_resources', 'reviewer_id');
 SELECT create_distributed_table('review_stages', 'organization_id');
-SELECT create_distributed_table('reviews', 'organization_id');
\ No newline at end of file
+SELECT create_distributed_table('reviews', 'organization_id');
+SELECT create_distributed_table('activities', 'organization_id');
```
aks30: @priyanka-TL Please add resource id as DC and make corresponding changes

aks30 (changes requested):
Reviewed on 29 Aug

src/controllers/v1/activities.js
priyanka-TL: added resource id as filter

src/database/models/actions.js
priyanka-TL: done. Removed hook

src/services/reviews.js
priyanka-TL: removed

src/distributionColumns.psql
priyanka-TL: yes done

src/validators/v1/activities.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * name : validators/v1/activities.js
+ * author : Priyanka Pradeep
+ * Date : 29-Aug-2024
+ * Description : Validations of activities controller
+ */
+
+module.exports = {
+	list: (req) => {
```
aks30, line 9: @priyanka-TL - Activity list will always be with resource id which is mandatory

src/database/models/reviews.js
```diff
@@ -55,5 +57,45 @@ module.exports = (sequelize, DataTypes) => {
 		}
 	)
 
+	Review.addHook('afterSave', async (instance, options) => {
+		try {
+			//get resource type
+			const resource = await sequelize.models.Resource.findOne(
```
aks30, line 63: @priyanka-TL why do we have to make one more query, won't the instance object have org id and type?

src/database/models/reviews.js
```diff
@@ -55,5 +57,45 @@ module.exports = (sequelize, DataTypes) => {
...
+				{
+					where: { id: instance.resource_id },
+					attributes: ['id', 'organization_id', 'type'],
+				},
+				{ raw: true }
+			)
+
+			if (resource?.id) {
```
aks30: @priyanka-TL in which case resource id won't be there

src/database/migrations/20240813123014-create-activities-table.js
```diff
@@ -0,0 +1,56 @@
...
+			},
+			updated_at: {
+				allowNull: false,
+				type: Sequelize.DATE,
+			},
+			deleted_at: {
+				type: Sequelize.DATE,
+			},
+		})
+		//add index
```
aks30, line 47: @priyanka-TL why do we need orgid and user id as index? just object ID is enough that is the only thing will be queries with.

src/services/projects.js
```diff
@@ -729,6 +725,14 @@ module.exports = class ProjectsHelper {
 			}
 
 			await resourceQueries.updateOne({ id: projectData.id }, resourcesUpdate)
+			//add user action
+			eventEmitter.emit(common.EVENT_ADD_USER_ACTION, {
```
aks30, line 729: @priyanka-TL why is this added here and not via hook ?

src/services/activities.js
```diff
@@ -0,0 +1,73 @@
...
+			const options = {
+				offset,
+				limit,
+			}
+
+			let filter = {
+				organization_id: orgId.toString(),
+			}
+
+			if (resourceId) {
```
aks30: @priyanka-TL if resource id is not given we should not query, without DC key the query would fail right?

src/services/actions.js
```diff
@@ -0,0 +1,183 @@
...
+	 */
+
+	static async delete(id) {
+		try {
+			const action = await actionQueries.findById(id)
+			if (!action) {
+				throw new Error('ACTION_NOT_FOUND')
+			}
+
+			const activities = await activitiesQueries.findAllActivities(
```
aks30, line 156: @priyanka-TL won't this query be non-optimal? Without DC and index?

aks30 (changes requested):
Reviewed on 29 Aug

src/services/actions.js
priyanka-TL: @aks30 we don't have another option here. When deleting the action, we can't include the org ID. We'll need to verify if any other organisation has activities linked to the action being deleted. Its deletion so its not a frequent activity.

src/services/activities.js
priyanka-TL: @aks30, I'm not fully convinced about adding object_id as DC. If we also include comments in the future, this might not be the best approach. Additionally, for the list, we won't be able to see all user actions for an organisation, which feels like a limitation.

However, I've made the changes as per your suggestion

src/services/projects.js
priyanka-TL: @aks30, as we've discussed before, once the user submits the resource for a second time or more, we can't identify that through the hook. The first submission sets the resource status to 'submitted,' but for subsequent submissions, it changes to 'in_review'.

src/database/migrations/20240813123014-create-activities-table.js
priyanka-TL: @aks30 previously i have designed like org admin can see all users actions, but as per the suggestion i have removed that. 
hence removing the primary key also

src/database/models/reviews.js
priyanka-TL: @aks30 just to reverify its added, i have removed that

src/database/models/reviews.js
priyanka-TL: @aks30, the reviews table doesn't include the resource type. We'll need to query the resource table to get the resource type.

src/validators/v1/activities.js
priyanka-TL: done

src/services/activities.js
```diff
@@ -0,0 +1,70 @@
...
+	 * @name list
+	 * @param {Integer} resourceId - The ID of the resource for which activities are listed.
+	 * @param {String} userId - Id of the logged in user
+	 * @param {Integer} page - Page number
+	 * @param {Integer} page - Page size limit
+	 * @param {String} search - Search text.
+	 * @returns {JSON} - activities list response.
+	 */
+
+	static async list(resourceId, userId, orgId, page, limit) {
```
aks30, line 25: @priyanka-TL where do we check if object id is not null?

src/database/models/reviews.js
aks30: @priyanka-TL In that case can we try where we are calling review save or update we set resource type as well, it might not save in DB but here in instance object we might get that.

aks30 (changes requested):
Added comments on 30 Aug

src/database/models/reviews.js
priyanka-TL: @aks30 After updating the reviews table we wont get type if we passed also

src/services/activities.js
priyanka-TL: @aks30 In validator we are checking

priyanka-TL: Added in service level too

PR #170, opened 2024-09-02
Scp route source

adithyadinesh0412:


src/constants/interface-routes/configs.json
```diff
@@ -0,0 +1,708 @@
...
+	"routes": [
+		{
+			"sourceRoute": "/scp/v1/permissions/list",
+			"type": "GET",
+			"priority": "MUST_HAVE",
+			"inSequence": false,
+			"orchestrated": false,
+			"targetPackages": [
+				{
+					"basePackageName": "self_creation_portal",
```
priyanka-TL: @adithyadinesh0412  we should use self-creation-portal right ?

priyanka-TL (changes requested):

PR #191, opened 2024-09-12
Integration test cases

priyanka-TL:


.circleci/config.yml
```diff
@@ -39,6 +39,30 @@ jobs:
...
+            echo "CLOUD_STORAGE_SECRET=$CLOUD_STORAGE_SECRET" >> ./src/.env
+            echo "CLOUD_STORAGE_BUCKETNAME=$CLOUD_STORAGE_BUCKETNAME" >> ./src/.env
+            echo "CLOUD_STORAGE_REGION=$CLOUD_STORAGE_REGION" >> ./src/.env
+            echo "CLOUD_ENDPOINT=$CLOUD_ENDPOINT" >> ./src/.env
+            ls -l ./src/.env # Check that the file is created
+            cat ./src/.env # Verify content for debugging
+
+
+      - run:
+          name: Replace User Service environment variables
```
adithyadinesh0412, line 64: Do we have to do this? As we are not using any cloud related services of user service? @priyanka-TL

src/integration-tests/certificates/certificates.spec.js
```diff
@@ -0,0 +1,16 @@
+const commonHelper = require('@commonTests')
```
adithyadinesh0412, line 1: How are we planning to execute certificate related scripts automatically here? @priyanka-TL  Can we add it in the circle ci config?

src/integration-tests/certificates/certificates.spec.js
```diff
@@ -0,0 +1,16 @@
...
+jest.setTimeout(10000)
+
+describe('Certificate APIs ', function () {
+	let userDetails
+	beforeAll(async () => {
+		userDetails = await commonHelper.logIn()
+	})
+
+	it('List Certificate Templates', async () => {
+		let res = await request.get('/scp/v1/certificates/list').query({ page: 1, limit: 10 })
```
adithyadinesh0412, line 12: can we have page and limit value in any common place? As it will be used through out the tests. @priyanka-TL

src/integration-tests/commonTests.js
```diff
@@ -2,30 +2,36 @@ var supertest = require('supertest') //require supertest
...
+		delay: 5000,
+		interval: 2500,
+		timeout: 100000,
+	}
+	await waitOn(opts)
+}
+
 const logIn = async () => {
 	try {
 		console.log('============>LOGIN 1 : ')
```
adithyadinesh0412: If we need console.log for any specific debugging, can we provide some meaningful statements inside? For example instead of 
============>LOGIN 1 :
we can have , 
============> ATTEMPTING LOGIN  :
and so on... 
@priyanka-TL

src/integration-tests/commonTests.js
```diff
@@ -58,14 +64,85 @@ const logIn = async () => {
...
+				.send({
+					title: 'content_creator',
+					user_type: 0,
+					organization_id: 1,
+					label: 'Content Creator',
+					visibility: 'PUBLIC'
+				})
+
+			if (createCreatorRole.statusCode != 200) {
+				console.log('Role Creation Failed')
```
adithyadinesh0412: can we give specific error message or the role title / code to better understand?  @priyanka-TL

src/integration-tests/config.js/responseSchema.js
```diff
@@ -0,0 +1,95 @@
+const listSchema = {
```
adithyadinesh0412: Can we add negative cases schema as well? for example here in this case if we attempt to call this api without token , api should throw 401 Unauthorised. Such cases can we handle? @priyanka-TL

src/integration-tests/modules.js/module.spec.js
```diff
@@ -0,0 +1,36 @@
+const commonHelper = require('@commonTests')
+const schema = require('./responseSchema')
+jest.setTimeout(10000)
+
+describe('Module APIs ', function () {
+	let userDetails
+	beforeAll(async () => {
+		userDetails = await commonHelper.logIn()
```
adithyadinesh0412: @priyanka-TL  this login function is creating a new user each time. can we change the approach? So that all the the test cases will be checked with a single user itself.

adithyadinesh0412 (changes requested):
Reviewed on 23rd Sept.

src/integration-tests/commonTests.js
```diff
@@ -58,14 +64,85 @@ const logIn = async () => {
...
+			})
+
+		if (existingReviewerRole.body.result?.data?.length == 0) {
+			let createReviewRole = await request.post('/user/v1/user-role/create')
+				.set(defaultHeaders)
+				.send({
+					title: 'reviewer',
+					user_type: 0,
+					organization_id: 1,
+					label: 'Content Creator',
```
adithyadinesh0412: Label should be corrected. @priyanka-TL

src/integration-tests/config.js/responseSchema.js
priyanka-TL: @adithyadinesh0412  as Akash said we can add negative test cases while developing

PR #196, opened 2024-09-19
delete draft comments

adithyadinesh0412:


src/controllers/v1/comments.js
```diff
@@ -42,12 +43,21 @@ module.exports = class comments {
...
-			const comment = await commentService.update(
-				req.params.id ? req.params.id : '',
-				req.query.resource_id,
-				req.body,
-				req.decodedToken.id
-			)
+			let comment
+			if (req.method === common.REQUEST_METHOD_DELETE) {
+				comment = await commentService.delete(
+					parseInt(req.params.id),
```
priyanka-TL, line 49: @adithyadinesh0412 do we need to convert to int? We can validate in the validator and throw error from there if its not a number

src/database/queries/comments.js
```diff
@@ -44,7 +44,37 @@ module.exports = class CommentData {
...
+	/**
+	 * comment delete
+	 * @method
+	 * @name delete
+	 * @param {Integer} id - comment id
+	 * @param {Integer} resourceId - resource id
+	 * @param {String} loggedInUserId - user id
+	 * @returns {JSON} - comment delete response.
+	 */
+	static async deleteOne(id, resourceId, loggedInUserId) {
```
priyanka-TL: @adithyadinesh0412  make this to userId

src/controllers/v1/comments.js
adithyadinesh0412: But the values are coming as string in the service and db is not able to find the right data. So I'm type casting it. @priyanka-TL

priyanka-TL (approved):

PR #203, opened 2024-10-06
Migration script

priyanka-TL:


src/scripts/migrateProjects.js
```diff
@@ -0,0 +1,520 @@
...
+const projectService = require('@services/projects')
+const entityService = require('@services/entities')
+const resourceService = require('@services/resource')
+const resourceQueries = require('@database/queries/resources')
+const _ = require('lodash')
+const MongoClient = require('mongodb').MongoClient
+// var ObjectId = require('mongodb').ObjectID
+
+//get mongo db url
+const mongoUrl = process.env.MONGODB_URL
```
adithyadinesh0412: can we add this in the `env.sample` also?
@priyanka-TL

src/scripts/migrateProjects.js
```diff
@@ -0,0 +1,520 @@
...
+
+if (!mongoUrl) {
+	throw new Error('MONGODB_URL is not set in the environment variables.')
+}
+
+// Path to the CSV file
+const outputPath = path.resolve(__dirname, 'migration_results.csv')
```
adithyadinesh0412: will the csv be able to handle if the result is volume is more? @priyanka-TL  
Should we go for a better approach?

package.json
```diff
@@ -1,6 +1,7 @@
 {
 	"dependencies": {
 		"@babel/parser": "^7.20.3",
-		"@babel/types": "^7.20.2"
+		"@babel/types": "^7.20.2",
+		"uuid": "^10.0.0"
```
aks30: @priyanka-TL why do we have 2 different package.json?

src/scripts/migrateProjects.js
```diff
@@ -0,0 +1,472 @@
+/**
+ * name : migrateProjects.js
+ * author : Priyanka Pradeep
+ * created-date : 24-Aug-2024
+ * Description : script to create the project from consumption side.
+ */
+
```
aks30: @priyanka-TL - Few questions?

We should use the org id of the creator for org id

Do we store the reference of mongo id in SCP?

Do we upload the JSON in cloud during migration?

aks30 (changes requested):
Reviewed on 14 Oct

src/scripts/migrateProjects.js
priyanka-TL: @aks30  Creator org Id is not available in template, Yes we are published_id. Yes uploading to cloud

package.json
priyanka-TL: Removed

PR #205, opened 2024-10-08
added migration for solution details

adithyadinesh0412:


priyanka-TL:
Already covered in submit for review

PR #227, opened 2024-10-21
Stages Implementation - Bug 1827

adithyadinesh0412:


priyanka-TL (approved):
Approved on Oct 21st, 11.52 AM

src/services/projects.js
```diff
@@ -245,6 +250,10 @@ module.exports = class ProjectsHelper {
 					updateData.is_under_edit = true
 				}
 
+				//get the resource stage
+				const stageData = await resourceService.getResourceStage(resourceId, orgId)
```
aks30: @priyanka-TL as discussed this is not required. STAGE will be set or modified only in creation, submit for review and publish

src/services/projects.js
```diff
@@ -726,7 +738,12 @@ module.exports = class ProjectsHelper {
 						status: common.REVIEW_STATUS_CHANGES_UPDATED,
 					}
 				)
-				resourceStatus = common.RESOURCE_STATUS_IN_REVIEW
+				// common.RESOURCE_STATUS_IN_REVIEW
+				const resourceFinalStatus = await resourceService.finalResourceStatus(
+					resourceId,
+					userDetails.organization_id
```
aks30: @priyanka-TL this should be resource org id

aks30: Change the first line of the function accordingly too

src/services/reviews.js
```diff
@@ -63,17 +63,26 @@ module.exports = class reviewsHelper {
 			}
 
 			// Update the status in the reviews table
+			await reviewsQueries.update(
+				{ id: review.id, organization_id: review.organization_id },
+				{ status: common.REVIEW_STATUS_REQUESTED_FOR_CHANGES }
+			)
+
+			const resourceStatus = await resourceService.finalResourceStatus(resourceId, resource.organization_id)
```
aks30: @priyanka-TL as discussed this is not required.

src/services/reviews.js
```diff
@@ -63,17 +63,26 @@ module.exports = class reviewsHelper {
...
+			let resourceUpdateObj = {
+				status: resourceStatus[resourceId],
+				last_reviewed_on: new Date(),
+			}
+
+			//update stage
+			let stageData = await resourceService.getResourceStage(resourceId, resource.organization_id)
```
aks30: @priyanka-TL as discussed this is not required.

src/services/resource.js
```diff
@@ -132,10 +113,39 @@ module.exports = class resourceHelper {
...
+				id: {
+					[Op.in]: uniqueResourceIds,
+				},
+				organization_id: {
+					[Op.in]: OrganizationIds,
+				},
+				status: common.REVIEW_STATUS_REQUESTED_FOR_CHANGES,
+			},
+			[
+				'id',
```
aks30: @adithyadinesh0412 confirm if we need to code all the column names

src/services/resource.js
```diff
@@ -1260,4 +1373,59 @@ module.exports = class resourceHelper {
...
 	}
+
+	/**
+	 * Get resources stage
+	 * @name getResourceStage
+	 * @param {String} organization_id - Org Id of the user
+	 * @param {String} resourceId - Resource Id
+	 * @returns {Object} - stage of the resource
+	 */
+	static async getResourceStage(resourceId, orgId) {
```
aks30: @priyanka-TL as discussed this is not required.

src/services/resource.js
```diff
@@ -226,13 +234,119 @@ module.exports = class resourceHelper {
...
 	}
+
+	/**
+	 * Return the mapping of resource Id and its final status
+	 * @name finalResourceStatus
+	 * @param {Array} resourceIds - Array of unique resource ids
+	 * @param {Array} organizationIds - Array of unique org ids
+	 * @returns {JSON} - List of resources with final status
+	 */
+	static async finalResourceStatus(resourceIds, organizationIds) {
```
aks30: @priyanka-TL Check if this and below function is really needed.

aks30 (changes requested):
Reviewed on 21 Oct

src/services/projects.js
adithyadinesh0412: removed

src/services/resource.js
adithyadinesh0412: only fetching the columns which is required

src/services/projects.js
adithyadinesh0412: resolved

src/services/reviews.js
adithyadinesh0412: removed

src/services/resource.js
adithyadinesh0412: removed

src/services/resource.js
adithyadinesh0412: yes its required, for the scenario which we discussed in morning

src/services/resource.js
```diff
@@ -226,13 +234,115 @@ module.exports = class resourceHelper {
...
+		const resourceReviews = await reviewsQueries.findAll(
+			{
+				organization_id: {
+					[Op.in]: organizationId,
+				},
+				resource_id: {
+					[Op.in]: resourceId,
+				},
+			},
+			['resource_id', 'status']
```
aks30: @priyanka-TL why we need resource id again its already there in function argument.

src/services/resource.js
```diff
@@ -226,13 +234,115 @@ module.exports = class resourceHelper {
...
+		let recourceStatusMapping = {}
+
+		// Get the review details of all the resources created by the logged-in user
+		const resourceReviews = await reviewsQueries.findAll(
+			{
+				organization_id: {
+					[Op.in]: organizationId,
+				},
+				resource_id: {
+					[Op.in]: resourceId,
```
aks30: @priyanka-TL this will never be multiple right, above also, using single eq operator

src/services/resource.js
```diff
@@ -226,13 +234,115 @@ module.exports = class resourceHelper {
...
+		return recourceStatusMapping
+	}
+
+	/**
+	 * Determine the final status of each resources - logic
+	 * @name determineResourceStatus
+	 * @param {Object} reviewDetails - Object of each review detail
+	 * @returns {JSON} - Final determined status of each reviews
+	 */
+	static async determineResourceStatus(reviewDetails) {
```
aks30: @priyanka-TL just change this function to take array of status and give a string of status

src/services/reviews.js
```diff
@@ -173,16 +175,15 @@ module.exports = class reviewsHelper {
...
-				resourceQueries.updateOne(
-					{ organization_id: resource.organization_id, id: resourceId },
-					{ status: common.RESOURCE_STATUS_IN_REVIEW }
-				),
-			])
+			await reviewsQueries.update(
+				{ id: review.id, organization_id: review.organization_id },
+				{ status: common.REVIEW_STATUS_INPROGRESS }
+			)
+			const resourceStatus = await resourceService.finalResourceStatus(resourceId, resource.organization_id)
```
aks30: @priyanka-TL why we need final status call, it will always be IN_PROGRESS if its submitted for review, else don't update something like that, check if we can avoid this call

aks30 (changes requested):
Reviewed on 22 Oct

src/services/resource.js
priyanka-TL: @adithyadinesh0412 can u please update here

src/services/reviews.js
priyanka-TL: @aks30 as we discussed in morning, consider there is 2 reviewer, and the reviewing parallel, one reviewer is alreday requested_for_change and then the resource status will be updated a requested_for_change. When the second person is start the review the resource status should not set as INPROGRESS hence the priority will be requested_for_change

src/services/resource.js
adithyadinesh0412: Removed `resource_id ` @aks30  , @priyanka-TL

src/services/resource.js
adithyadinesh0412: Yes @aks30  , Updated the code

src/services/resource.js
adithyadinesh0412: @aks30 Updated to return a single string from array of status.

src/services/resource.js
```diff
@@ -132,10 +113,39 @@ module.exports = class resourceHelper {
 			page,
 			limit
 		)
+		const requestedForChangesResources = await resourceQueries.resourceList(
```
aks30: @adithyadinesh0412 as discussed just get the count and don't need projection

src/services/reviews.js
aks30: Please connect with @adithyadinesh0412 i explained how we can potentially do it

aks30 (changes requested):
Reviewed on 22 Oct 7:45 PM

PR #248, opened 2024-11-14
permission issue fixed

adithyadinesh0412:


src/services/permissions.js
```diff
@@ -66,7 +66,11 @@ module.exports = class PermissionsHelper {
 			const filter = { id }
 			const permissions = await permissionsQueries.findPermissionById(id)
 			if (!permissions) {
-				throw new Error('PERMISSION_NOT_FOUND')
+				return responses.failureResponse({
```
priyanka-TL, line 69: @adithyadinesh0412 just do this return it catch and throw the message from here

src/services/permissions.js
```diff
@@ -88,7 +88,11 @@ module.exports = class PermissionsHelper {
 				})
 			}
 		} catch (error) {
-			throw error
+			return responses.failureResponse({
+				message: error.message,
```
priyanka-TL: @adithyadinesh0412  can u please add message: error.message || error

priyanka-TL (approved):

src/services/permissions.js
```diff
@@ -88,7 +88,11 @@ module.exports = class PermissionsHelper {
 				})
 			}
 		} catch (error) {
-			throw error
+			return responses.failureResponse({
```
aks30: @adithyadinesh0412 what if DB is down, will you still give bad request?

aks30 (changes requested):
Reviewed on 15 Nov

src/services/permissions.js
aks30: We should have different response based on actual issue, bad request only if frontend is giving invalid data

adithyadinesh0412: Resolved this. 
sending back error response if permission is not found.
For any other issues , catch block with send out generic error "Oops something went wrong" with 500 status code

priyanka-TL (approved):
Approved on Nov 19

PR #256, opened 2024-11-25
updatePermissions API fix - 2027

priyanka-TL:


adithyadinesh0412 (approved):
Reviewed on Nov 25 - 1.24PM

src/services/permissions.js
```diff
@@ -92,7 +92,15 @@ module.exports = class PermissionsHelper {
 				})
 			}
 		} catch (error) {
-			throw error
+			if (error.name === common.SEQUELIZE_DATABASE_ERROR && error.parent?.code === '22003') {
```
aks30: @priyanka-TL can we not check whether its integer or not in validator only?

src/services/permissions.js
```diff
@@ -125,7 +133,15 @@ module.exports = class PermissionsHelper {
 				result: {},
 			})
 		} catch (error) {
-			throw error
+			if (error.name === common.SEQUELIZE_DATABASE_ERROR && error.parent?.code === '22003') {
```
aks30: @priyanka-TL can we not check whether its integer or not in validator only?

aks30 (changes requested):
Reviewed on 25 Nov

src/services/permissions.js
priyanka-TL: done

PR #258, opened 2024-11-26
Permission rollout

priyanka-TL:


src/constants/common.js
```diff
@@ -195,5 +195,4 @@ module.exports = {
 	RECOMMENDED_DURATION: 'recommended_duration',
 	NUMBER: 'number',
 	STRING: 'string',
-	SEQUELIZE_DATABASE_ERROR: 'SequelizeDatabaseError',
```
adithyadinesh0412, line 198: Why was this removed here? @priyanka-TL . I think this was added as a bug fix right? is this change relevant here?

adithyadinesh0412 (changes requested):
Reviewed on 28th Nov - 1.45PM

adithyadinesh0412 (approved):
Approved on 28 Nov - 2.05PM

PR #262, opened 2024-11-27
Rollout create

priyanka-TL:


src/services/rollouts.js
```diff
@@ -0,0 +1,136 @@
...
+	 * Rollout create
+	 * @method
+	 * @name create
+	 * @param {Object} req - request data.
+	 * @returns {JSON} - project id
+	 */
+	static async create(bodyData, loggedInUserId, orgId) {
+		try {
+			//validate the title length
+			const isTitleInvalid = utils.validateTitle(bodyData.title)
```
adithyadinesh0412: Can we do this check as part of validator? @priyanka-TL

src/services/rollouts.js
```diff
@@ -0,0 +1,136 @@
...
+			if (isTitleInvalid) {
+				return responses.failureResponse({
+					message: 'CHARACTER_LIMIT_EXCEED',
+					statusCode: httpStatusCode.bad_request,
+					responseCode: 'CLIENT_ERROR',
+				})
+			}
+
+			//validation for start date and end date
+			if (bodyData.start_date && bodyData.end_date) {
```
adithyadinesh0412: This also can we do it in validator? So that these basic checks are validated before we reach service file @priyanka-TL

src/services/rollouts.js
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			let rolloutData = {
+				title: bodyData.title,
+				resource_type: resource.type,
+				resource_id: resource.id,
+				status: common.ROLLOUT_STATUS_PENDING,
+				start_date: bodyData.start_date,
+				end_date: bodyData.end_date,
+				type: common.ROLLOUT_TYPE_PROGRAM,
```
adithyadinesh0412, line 43: We should check the type of the resource and assign this type field. @priyanka-TL

src/services/rollouts.js
```diff
@@ -0,0 +1,136 @@
...
+
+					if (updateCount === 0) {
+						return responses.failureResponse({
+							message: 'ROLLOUT_NOT_FOUND',
+							statusCode: httpStatusCode.bad_request,
+							responseCode: 'CLIENT_ERROR',
+						})
+					}
+				} else {
+					throw new Error('FILE_UPLOADED_FAILED')
```
adithyadinesh0412, line 100: should we rollback the entry made to rollout table in case of a file upload failure? @priyanka-TL

src/validators/v1/rollouts.js
```diff
@@ -0,0 +1,42 @@
...
+			.trim()
+			.optional({ checkFalsy: true })
+			.notEmpty()
+			.withMessage('id param is empty')
+			.isNumeric()
+			.withMessage('id param is invalid, must be an integer')
+			.isInt({ min: 1, max: 2147483647 })
+			.withMessage('Id is not valid')
+
+		req.checkParams('start_date').trim().notEmpty().withMessage('start_date param is empty')
```
adithyadinesh0412: @priyanka-TL  start_date is passed in the body right?  not as a param. Please recheck this file.

adithyadinesh0412 (changes requested):
Reviewed on 28th Nov - 1.42PM

src/services/rollouts.js
priyanka-TL: type is not the resource type its rollout type

adithyadinesh0412 (approved):

PR #281, opened 2024-12-10
Configure token prefix

priyanka-TL:


src/services/organization-extension.js
```diff
@@ -270,6 +270,7 @@ module.exports = class orgExtensionsHelper {
 				instance: {
 					auto_save_interval: utils.convertToInteger(process.env.RESOURCE_AUTO_SAVE_TIMER),
 					note_length: utils.convertToInteger(process.env.MAX_RESOURCE_NOTE_LENGTH),
+					is_auth_token_bearer: process.env.IS_AUTH_TOKEN_BEARER === 'true',
```
aks30, line 273: @priyanka-TL Do we need this since anyways default value will be set right?

src/middlewares/authenticator.js
```diff
@@ -45,6 +45,7 @@ module.exports = async function (req, res, next) {
 	try {
 		let roleValidation = false
 		let decodedToken
+		const isBearerRequired = process.env.IS_AUTH_TOKEN_BEARER === 'true'
```
aks30, line 48: @priyanka-TL Do we need this since anyways default value will be set right?

aks30 (changes requested):
Reviewed on 26 Dec

src/services/organization-extension.js
priyanka-TL: @aks30  based on this key frontend can decide they need to pass bearer or not

src/middlewares/authenticator.js
priyanka-TL: @aks30  based on this key, we written the logic to get the token, like remove bearer if IS_AUTH_TOKEN_BEARER is false. Same logic that mentoring follows

PR #290, opened 2024-12-16
Resource publish

priyanka-TL:


src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+ * @name formatTemplate
+ * @param {Object} templateData - Project template data
+ * @returns {Object} - Response contains formatted template
+ */
+const formatTemplate = (templateData) => {
+	try {
+		let template = {
+			title: templateData.title,
+			description: templateData.objective || '',
+			keywords: templateData.keywords ? templateData.keywords.split(',').map((k) => k.trim()) : [],
```
adithyadinesh0412: Can we handle this scenario? 
If keywords is coming as an array it should not break. As we don't have any validation for keyword. I think that is a valid scenario. 
@priyanka-TL

src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+				: [],
+			createdBy: templateData.user_id,
+			updatedBy: templateData.user_id,
+			learningResources: templateData.learning_resources ? convertResources(templateData.learning_resources) : [],
+			isReusable: true,
+			taskSequence: [], // Initially empty
+			deleted: false,
+			status: common.PUBLISHED_STATUS,
+			externalId: generateExternalId(templateData.title),
+			entityType: '',
```
adithyadinesh0412, line 146: entityType should be `school` by default for Projects. Please check with @VISHNUDAS-tunerlabs .
@priyanka-TL

src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+			metaInformation: {
```
adithyadinesh0412: Can we add goal , rationale etc... from the template data? Just in case in future we have those fields coming from SCP we need not make change here.  
`
templateDetails?.goal ? templateDetails?.goal  : ''
`

@priyanka-TL

src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+				formattedName: formatCategoriesName(category.value),
+				externalId: category.value.replace(/_/g, '').toLowerCase(),
+			}
+		})
+
+		// Fetch existing categories by externalId
+		const existingCategories = await categoriesCollection
+			.find({ externalId: { $in: formattedCategories.map((cat) => cat.externalId) } })
+			.toArray()
+		const existingExternalIds = new Set(existingCategories.map((cat) => cat.externalId))
```
adithyadinesh0412: Can we convert it to array? 
`[ ...new Set(existingCategories.map((cat) => cat.externalId))]`
or are we intending to use it as Set only? 
Because Arrays tends to be faster than Sets for smaller datasets. If it is a major code code change let's not do that because I believe the change should bring only a minor performance change per it's say. 
@priyanka-TL

src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+	try {
+		const taskCollection = mongoDb.collection(COLLECTIONS.TASKS)
+		const taskIds = []
+		const externalIds = []
+
+		for (const task of tasks) {
+			// Format the task data
+			const taskData = {
+				name: task.name,
+				description: task.name,
```
adithyadinesh0412, line 271: Can we have `task?.description  ? task?.description  : ''`  In case of an adopter needs description for task.  @priyanka-TL

src/requests/consumption.js
```diff
@@ -0,0 +1,377 @@
...
+				externalId: generateExternalId(task.name),
+				type: task.type,
+				isDeleted: !task.is_mandatory,
+				isDeletable: !task.is_mandatory,
+				sequenceNumber: task.sequence_no,
+				projectTemplateId: templateId,
+				projectTemplateExternalId: templateExternalId,
+				hasSubTasks: !!task.children?.length,
```
adithyadinesh0412: Can we write it `task.children?.length > 0` for better readability ? and eliminate a second degree negation. @priyanka-TL

adithyadinesh0412 (changes requested):
Reviewed on 23rd Dec 7.05 PM

src/requests/consumption.js
priyanka-TL: @adithyadinesh0412 If the key is different in scp, again we need to make a code change, i think we can update this later

src/requests/consumption.js
priyanka-TL: @priyanka-TL  we dont have description key in scp. desription we are storing in scp as name, and also name = description as of now. Instaed of modifying each task unnecessory, we can add name and descriptio both same as of now, if they want this different, let scp need to handle that first

src/requests/consumption.js
priyanka-TL: @adithyadinesh0412  by default it will be empty,

src/envVariables.js
```diff
@@ -204,6 +204,24 @@ let environmentVariables = {
 		optional: true,
 		default: 'rollout_manager',
 	},
+	CONSUMPTION_SERVICE_BASE_URL: {
+		message: 'Default data managers required.',
```
aks30: @priyanka-TL - Please correct the message.

src/generics/utils.js
```diff
@@ -612,6 +612,77 @@ function isEnglish(text) {
...
+	if (typeof keywords === 'string') return keywords.split(',').map((k) => k.trim())
+	return []
+}
+
+/**
+ * Format meta-information
+ * @param {Object} templateData - Template data
+ * @returns {Object} - Meta-information
+ */
+function formatMetaInformation(templateData) {
```
aks30: @priyanka-TL Rename it to formatProjectMetaInformation

aks30 (changes requested):
Reviewed on 26 Dec

src/envVariables.js
priyanka-TL: Done

src/generics/utils.js
priyanka-TL: done

PR #329, opened 2025-01-07
Story 1356 consumption

adithyadinesh0412:


src/.env.sample
```diff
@@ -197,3 +197,6 @@ CONSUMPTION_SERVICE_BASE_URL=/project/
 
 #Indicate If auth token is bearer or not
 IS_AUTH_TOKEN_BEARER=false
+
+# consumption side pre-signed url generation end-point
+CONSUMPTION_SERVICE_PRESIGNED_URL='/project/v1/cloud-services/files/preSignedUrls'
```
priyanka-TL: @adithyadinesh0412 do we have any key to store consumption service base name. If there then we can store only the end point here

src/controllers/v1/rollouts.js
```diff
@@ -122,6 +122,7 @@ module.exports = class rollouts {
 		try {
 			const rollout = await rolloutService.publish(
 				req.params.id,
+				req.userToken,
```
priyanka-TL: @adithyadinesh0412 add that in the end

src/envVariables.js
```diff
@@ -237,6 +237,16 @@ let environmentVariables = {
 		optional: true,
 		default: true,
 	},
+	CONSUMPTION_SERVICE_PRESIGNED_URL: {
+		message: 'Default consumption side pre-signed url for rollout publish required',
```
priyanka-TL: @adithyadinesh0412 if consumption service is not self then only we require this key right

src/requests/consumption.js
```diff
@@ -921,23 +922,29 @@ async function createSvg(certificateData, loggedInUserId) {
...
 				},
-				ref: common.CERTIFICATE,
 			}
 			// generate signed url
-			const getSignedUrl = await filesService.getSignedUrl(payloadData, common.CERTIFICATE, loggedInUserId, false)
-			if (!getSignedUrl.result) {
-				throw new Error('FAILED_TO_GENERATE_SIGNED_URL')
+			// const getSignedUrl = await filesService.getSignedUrl(payloadData, common.CERTIFICATE, loggedInUserId, false)
+			const headers = {
+				'internal-access-token': process.env.INTERNAL_ACCESS_TOKEN,
```
priyanka-TL: @adithyadinesh0412 we don't require internal access token hence it is not added under internal access token in project service

src/requests/consumption.js
```diff
@@ -921,23 +922,29 @@ async function createSvg(certificateData, loggedInUserId) {
...
+				'X-auth-token': userToken,
 			}
-
-			if (!getSignedUrl.result) {
+			const getSignedUrl = await generateConsumptionPresignedUrl(
```
priyanka-TL: @adithyadinesh0412 can we rename to generatePresignedUrlInConsumption

src/requests/consumption.js
```diff
@@ -1057,10 +1090,26 @@ async function insertCertificateTemplate(certificateData, solutionId, programId,
...
+		{
+			$set: {
+				certificateTemplateId: result.insertedId,
+			},
+		}
+	)
+
+	// Validate the result of the template creation
+	if (!resultUpdateProjecTemplate) {
+		throw new Error(`Failed to update the certificate template into the ${COLLECTIONS.TEMPLATES} collection.`)
```
priyanka-TL: @adithyadinesh0412 check the updated count, the response won't be empty anyways

priyanka-TL (changes requested):
Reviewed on 8.53 PM

src/envVariables.js
adithyadinesh0412: Yes. changing the message

PR #342, opened 2025-01-20
Story-1077 - Create model, migrations and Role permission mapping

priyanka-TL:


src/distributionColumns.psql
```diff
@@ -14,4 +14,5 @@ SELECT create_distributed_table('activities', 'object_id');
 SELECT create_distributed_table('organization_configs', 'organization_id');
 
 
-SELECT create_distributed_table('rollouts', 'organization_id');
\ No newline at end of file
+SELECT create_distributed_table('rollouts', 'organization_id');
+SELECT create_distributed_table('program_resource_mapping', 'program_id');
```
adithyadinesh0412: Distribution on program_id or organization_id? Not very clear about this @priyanka-TL . Can we discuss?

priyanka-TL: Program id, is better because the node will contain less number of data, for listing we may need to look into different nodes if we need to add organization id. The purpose of this table is to get the mapping resource right, so program id can be an optimized way to get data. Please let me know if you have any other findings we can discuss

priyanka-TL: Also query filters by program_id, distributing the table by program_id ensures that all rows for a specific program_id are stored on the same node. This minimizes cross-node queries and improves performance.

priyanka-TL: Checked that we can add organization_id as DC , hence too many node creation causes citus level issues.

adithyadinesh0412 (approved):
Approved on : 21 - Jan - 2025

PR #350, opened 2025-01-29
Add resource to Program

adithyadinesh0412:


src/constants/blacklistConfig.js
```diff
@@ -141,6 +141,25 @@ const programs = {
 		'created_at',
 		'updated_at',
 	],
+	addResources: [
```
priyanka-TL: @adithyadinesh0412  Its handling in validator please remove this

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
 			throw error
 		}
 	}
+
+	static async addResources(programId, updateBody, loggedInUserId, orgId) {
```
priyanka-TL: @adithyadinesh0412  add function sigature

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+		try {
+			// combine all ids to fetch from resources table
+			const allResources = [programId, ...updateBody.resource_ids.map((resourceId) => parseInt(resourceId))]
```
priyanka-TL: @adithyadinesh0412  Can we rename to resourceIds

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+			// Fetch the program to be updated
+			const fetchProgram = await resourceQueries.findAll({
```
priyanka-TL: @adithyadinesh0412  instead of fetchProgram it can be resources hence fetching all the resources at once right

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+				id: { [Op.in]: allResources },
+				organization_id: orgId,
```
priyanka-TL: @adithyadinesh0412 Here we cant find with org id. Because in browse existing it can be list of all published resource under user org or all org,

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+			})
+
+			let programDetails,
```
priyanka-TL: @adithyadinesh0412  we should validate the programId exist in resources

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+				resourceToCreate = []
+
+			fetchProgram.forEach((resource) => {
```
priyanka-TL: @adithyadinesh0412 we can remove that resource from the fetchProgram then loop only the req body resources

src/services/programs.js
```diff
@@ -236,6 +235,77 @@ module.exports = class ProgramsHelper {
...
+				// find program from the resources
+				if (resource.id == programId && resource.type == common.RESOURCE_TYPE_PROGRAM) {
+					if (resource.created_by != loggedInUserId) {
```
priyanka-TL: @adithyadinesh0412  this also can be moved to top

priyanka-TL (changes requested):
Reviewed on Jan 29

src/services/programs.js
adithyadinesh0412: We are doing that @priyanka-TL 
`if (!programDetails?.id) {
				throw new Error('PROGRAM_NOT_FOUND')
			}`

src/services/programs.js
adithyadinesh0412: Handling that inside the loop and segregating only the resources to create in another variable, @priyanka-TL

src/services/programs.js
adithyadinesh0412: Here since we have all the resources in a single variable , we can do it only here. 
Program add resource can only be done by the creator himself. 
But other resource owner can be different.
@priyanka-TL

src/services/programs.js
priyanka-TL: Can we do this above

src/services/programs.js
```diff
@@ -236,6 +235,86 @@ module.exports = class ProgramsHelper {
...
+	static async addResources(programId, updateBody, loggedInUserId, orgId) {
+		try {
+			// combine all ids to fetch from resources table
+			const resourceIds = [programId, ...updateBody.resource_ids.map((resourceId) => parseInt(resourceId))]
+			// Fetch the program to be updated
+			const fetchProgramAndResources = await resourceQueries.findAll({
+				id: { [Op.in]: resourceIds },
+			})
+
+			let programDetails,
```
priyanka-TL: @adithyadinesh0412  find the program details here itself

  const programDetails = fetchProgramAndResources.find(
            (resource) => resource.id === programId && resource.type === common.RESOURCE_TYPE_PROGRAM
        );

        if (!programDetails || programDetails.created_by !== loggedInUserId) {
            throw new Error('PROGRAM_NOT_FOUND_OR_UNAUTHORIZED');
        }

src/services/programs.js
```diff
@@ -236,6 +235,86 @@ module.exports = class ProgramsHelper {
...
+	 * @param {string} programId - resource id of the program to add resources
+	 * @param {Object} bodyData - Request body data.
+	 * @param {string} loggedInUserId - The ID of the logged-in user.
+	 * @param {string} orgId - The ID of the organization.
+	 * @returns {JSON} - Program ID or error response.
+	 */
+	static async addResources(programId, updateBody, loggedInUserId, orgId) {
+		try {
+			// combine all ids to fetch from resources table
+			const resourceIds = [programId, ...updateBody.resource_ids.map((resourceId) => parseInt(resourceId))]
```
priyanka-TL, line 436: @adithyadinesh0412  this is the optimized code please check
 static async addResources(programId, updateBody, loggedInUserId, orgId) {
    try {
        // Validate input
        if (!programId || !updateBody?.resource_ids || !Array.isArray(updateBody.resource_ids)) {
            throw new Error('INVALID_INPUT');
        }

        // Combine all IDs to fetch from the resources table
        const resourceIds = [programId, ...updateBody.resource_ids.map((resourceId) => parseInt(resourceId))];

        // Fetch all resources in a single query
        const fetchProgramAndResources = await resourceQueries.findAll({
            id: { [Op.in]: resourceIds },
        });

        // Early validation: Check if the program exists and belongs to the logged-in user
        const programDetails = fetchProgramAndResources.find(
            (resource) => resource.id === programId && resource.type === common.RESOURCE_TYPE_PROGRAM
        );

        if (!programDetails || programDetails.created_by !== loggedInUserId) {
            throw new Error('PROGRAM_NOT_FOUND_OR_UNAUTHORIZED');
        }

        // Segregate reusable and non-reusable resources
        const resourceToCreate = [];
        for (const resource of fetchProgramAndResources) {
            if (resource.id === programId) continue; // Skip the program itself

            if (resource.type !== common.RESOURCE_TYPE_PROGRAM) {
                if (resource.is_reusable) {
                    resourceToCreate.push(resource);
                } else {
                    throw new Error('FORBIDDEN_RESOURCE_IN_PROGRAM');
                }
            }
        }

        // Create copies of reusable resources in parallel
        let createdResources = [];
        if (resourceToCreate.length > 0) {
            createdResources = await handleResources(resourceToCreate, programId, orgId, loggedInUserId);
        }

        // Fetch program details and prepare data for upload
        const fetchProgramDetails = await resourceService.getDetails(programId, orgId);
        const programData = _.omit(fetchProgramDetails.result, [
            'created_at',
            'updated_at',
            'created_by',
            'updated_by',
            'status',
            'organization',
        ]);

        // Upload and update the program resource
        await uploadAndUpdateResource(
            programId,
            orgId,
            loggedInUserId,
            programData,
            common.PROGRAM_UPLOAD_FILE_NAME,
            common.RESOURCE_TYPE_PROGRAM
        );

        return responses.successResponse({
            statusCode: httpStatusCode.ok,
            message: 'RESOURCE_ADDED_TO_PROGRAM',
            result: programId,
        });
    } catch (error) {
        // Log the error for debugging
        console.error('Error in addResources:', error.message);
        throw error; // Re-throw the error for the caller to handle
    }
}

priyanka-TL (changes requested):
REviewed on Jan 29 3.25 PM

src/services/programs.js
adithyadinesh0412: I think this will add one more loop to the function. so i was thinking to handle this in the one loop we already have. @priyanka-TL

PR #369, opened 2025-02-11
circular dependancy fixed

adithyadinesh0412:


src/services/programs.js
```diff
@@ -75,7 +75,7 @@ module.exports = class ProgramsHelper {
 			try {
 				// Upload program to cloud
 				if (programId) {
-					await utils.uploadAndUpdateResource(
+					await resourceService.uploadAndUpdateResource(
```
priyanka-TL, line 79: @adithyadinesh0412 require reviewsResourcesQueries in program service

PR #371, opened 2025-02-13
Copying and editing program

priyanka-TL:
[Story: https://katha.shikshalokam.org/story-view-1252.html](https://katha.shikshalokam.org/story-view-1252.html)

### DIFF
--- a/src/services/programs.js
+++ b/src/services/programs.js
@@ -26,8 +26,45 @@ module.exports = class ProgramsHelper {
 	 * @param {string} orgId - The ID of the organization.
 	 * @returns {JSON} - Program ID or error response.
 	 */
-	static async create(bodyData, loggedInUserId, orgId) {
+	static async create(bodyData, loggedInUserId, orgId, referenceId = null) {
 		try {
+			if (referenceId) {
+				// check if the reference project Id is valid or not
+				const referenceProject = await resourceQueries.findOne({
+					id: referenceId,
+					status: common.RESOURCE_STATUS_PUBLISHED,
+					stage: common.RESOURCE_STAGE_COMPLETION,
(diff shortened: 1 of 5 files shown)


src/services/programs.js
```diff
@@ -25,10 +25,65 @@ module.exports = class ProgramsHelper {
...
+								'updated_by',
+								'submitted_on',
+								'published_on',
+								'last_reviewed_on',
+								'is_under_edit',
+							],
+						},
+					}
+				)
+
```
adithyadinesh0412, line 64: before calling for programDetails can we check if the referenceProject is present or not?  
if(referenceProject.id) {
throw error
}
so that in case of invalid reference id , it need not call program detail

src/validators/v1/programs.js
```diff
@@ -34,6 +34,14 @@ module.exports = {
 			.withMessage('id param is invalid, must be an integer')
 			.isInt({ min: 1, max: 2147483647 })
 			.withMessage('Id is not valid')
+
+		req.checkParams('reference_id')
```
adithyadinesh0412, line 38: can we add 
.isInt({ min: 1, max: 2147483647 })
validation as well? @priyanka-TL

adithyadinesh0412 (changes requested):
Reviewed on 17th Feb. 
Please include integration test  as well.

priyanka-TL:
For the create api we already have the integration test. To test the copy flow, we can do once the program publish part is completed

adithyadinesh0412 (approved):
Sure @priyanka-TL  , please note down this as an action item in our tracker where we added integration as an action item so that we don't miss out. 
Approved : 17th Feb 11.15AM

PR #407, opened 2025-02-28
buf fix and optimize the updating fields

priyanka-TL:
https://katha.shikshalokam.org/bug-view-2457.html
https://katha.shikshalokam.org/bug-view-2504.html


adithyadinesh0412 (approved):
Approved on 13th March

src/constants/blacklistConfig.js
```diff
@@ -140,6 +145,38 @@ const programs = {
 		'next_stage',
 		'created_at',
 		'updated_at',
+		'is_under_edit',
+		'deleted_at',
+		'metaData',
+		'meta',
+		'stage',
+		'organizations',
+		'is_comments',
```
aks30, line 154: @priyanka-TL what is this key for?

src/constants/blacklistConfig.js
```diff
@@ -140,6 +145,38 @@ const programs = {
...
+		'published_id',
+		'reference_id',
+		'next_stage',
+		'created_at',
+		'updated_at',
+		'is_under_edit',
+		'deleted_at',
+		'meta',
+		'is_comments',
+		'formMeta',
```
aks30, line 177: @priyanka-TL what is the diff between meta and fromMeta?

priyanka-TL: @aks30  its  the keys frontend is not removed from their end hence we are removing

src/constants/blacklistConfig.js
priyanka-TL: is_comments is the key that we are sending to identify the resources in the program have comments or not, that need not to be stored in the blob, It will add based on the comments.

src/constants/blacklistConfig.js
aks30: But why we have 2 different keys, what data gets stored in each?

src/constants/blacklistConfig.js
aks30: Can we rename it has_open_comments?

src/constants/blacklistConfig.js
priyanka-TL: FormMeta frontend is used for their processing, we requested to remove that. Hence that is not done from their end we are removing now. Meta is used by the backend, we are storing start date end date

src/constants/blacklistConfig.js
priyanka-TL: Hence listing also we are using the same key, we need to inform the frontend team also, that can we taken up in the next release?

PR #412, opened 2025-03-14
Scp program publish

adithyadinesh0412:


src/services/programs.js
```diff
@@ -956,7 +956,7 @@ module.exports = class ProgramsHelper {
 				{
 					user_id: userDetails.id,
 					resource_id: {
-						[Op.in]: [programId, ...resourceIds],
+						[Op.in]: [parseInt(programId), ...resourceIds],
```
priyanka-TL: @adithyadinesh0412  can we do this from the controller itself

src/services/rollouts.js
```diff
@@ -14,6 +14,7 @@ const { Op } = require('sequelize')
 const kafkaCommunication = require('@generics/kafka-communication')
 const entityModelMappingQuery = require('@database/queries/entityModelMapping')
 const utils = require('@generics/utils')
+const reviewService = require('@services/reviews')
```
priyanka-TL: @adithyadinesh0412  where are we using reviews

priyanka-TL (changes requested):
REviewed on Mar 14 th 5 PM

PR #420, opened 2025-03-20
Fix for delete a newly added resource inside already published program.

priyanka-TL:
https://katha.shikshalokam.org/bug-view-2556.html

### DIFF
--- a/src/services/programs.js
+++ b/src/services/programs.js
@@ -654,7 +654,7 @@ module.exports = class ProgramsHelper {
 			}
 
 			// Check if the program is published
-			if (program.status === common.PUBLISHED_STATUS || program.published_id) {
+			if (program.status === common.PUBLISHED_STATUS || program.published_id || program.published_on) {
 				// Fetch all resources linked to the program
 				const mappedResources = await programResourceMappingQueries.findAll({
 					program_id: programId,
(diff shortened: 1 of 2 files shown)


src/services/programs.js
```diff
@@ -654,12 +654,49 @@ module.exports = class ProgramsHelper {
...
+					return responses.failureResponse({
+						message: 'INVALID_RESOURCES_FOUND',
+						statusCode: httpStatusCode.bad_request,
+						responseCode: 'CLIENT_ERROR',
+					})
+				}
+
+				// If the program is published, only allow removing resources added after publishing
+				const resourcesAddedBeforePublishing = bodyData.resource_ids.filter((resourceId) => {
+					const resourceAddedDate = existingResourceMap[resourceId]
```
adithyadinesh0412: Please do the check directly. @priyanka-TL

adithyadinesh0412 (changes requested):

src/services/programs.js
```diff
@@ -654,12 +654,49 @@ module.exports = class ProgramsHelper {
...
 
 			// Check if the program is published
-			if (program.status === common.PROGRAM_STATUS_PUBLISHED || program.published_id) {
-				return responses.failureResponse({
-					message: 'CANNOT_REMOVE_RESOURCE_FROM_PUBLISHED_PROGRAM',
-					statusCode: httpStatusCode.bad_request,
-					responseCode: 'CLIENT_ERROR',
+			if (program.status === common.PUBLISHED_STATUS || program.published_id || program.published_on) {
+				// Fetch all resources linked to the program
+				const mappedResources = await programResourceMappingQueries.findAll({
```
adithyadinesh0412, line 659: We just need `created_at` and `id` data right? @priyanka-TL . Pass attribute to limit the data fetch

adithyadinesh0412:
Reviewed on 20th March

src/services/programs.js
priyanka-TL: no we need resource_id also id will be different

adithyadinesh0412: yes, fetching only keys which are needed can improve performance. It will have performance impact better that a select * query. @priyanka-TL

adithyadinesh0412 (approved):
Approved on 20th March

PR #459, opened 2025-04-24
consumption side changes

adithyadinesh0412:


src/constants/common.js
```diff
@@ -236,6 +236,7 @@ module.exports = {
 	},
 	PUBLISHED_STATUS: 'published',
 	CONSUMPTION_SERVICE_SELF: 'self',
+	SUNBIRD: 'sunbird',
```
priyanka-TL, line 239: @adithyadinesh0412 can you replace Diksha with sunbird, where ever we are using diksha use sunbird instead

priyanka-TL (changes requested):

src/constants/common.js
adithyadinesh0412: Done @priyanka-TL

priyanka-TL (approved):

PR #500, opened 2025-09-24
Migration program

priyanka-TL:
## Summary by CodeRabbit

* **New Features**
  * Added a searchable API endpoint for finding entity types by query; endpoint is configurable via environment settings.

* **Chores**
  * Admin/migration utilities now include configurable settings (entity-type keys, batching, caching, timeouts, exclusions) for more reliable migrations.

* **Enhancement**
  * Program/solution migration now preserves tenant scope so migrated items are created with tenant-specific context.

### DIFF
--- a/src/constants/endpoints.js
+++ b/src/constants/endpoints.js
@@ -19,4 +19,7 @@ module.exports = {
 	FIND_ENTITIES_BY_QUERY: process.env.FIND_ENTITIES_BY_QUERY_ENDPOINT
 		? process.env.FIND_ENTITIES_BY_QUERY_ENDPOINT
 		: 'v1/entities/find',
+	ENTITY_TYPES_FIND_BY_QUERY: process.env.ENTITY_TYPES_FIND_BY_QUERY_ENDPOINT
+		? process.env.ENTITY_TYPES_FIND_BY_QUERY_ENDPOINT
+		: 'v1/entityTypes/find',
 }
--- a/src/scripts/elevateProject/migrateProgramsAndSolutions.js
+++ b/src/scripts/elevateProject/migrateProgramsAndSolutions.js
@@ -1,5 +1,5 @@
 /**
- * name : migrateProgram.js
+ * name : migrateProgramsAndSolutions.js
  * author : Priyanka Pradeep
  * created-date : 24-Aug-2024
  * Description : Script to migrate programs and solutions with tenant support (Refactored)
@@ -122,7 +122,7 @@ const CURSOR_TIMEOUT = 30 * 60 * 1000 // 30 minutes cursor timeout
 					components: { $exists: true, $type: 'array', $not: { $size: 0 } },
 					tenantId: { $nin: [null, ''] },
 					orgId: { $nin: [null, ''] },
-					_id: ObjectId('682c3e4076d1f500145763cf'),
+					_id: ObjectId('682c147bba875600144d91bb'),
 				},
 			},
 			{
@@ -407,6 +407,9 @@ async function processProgram(
 
 	// generate targeting criteria for program scope
 	let programTargetingCriteriaRes = await generateTargetingCriteria(program.scope, tenant_code, validateScopeRes)
+	console.log(JSON.stringify(program?.scope, null, 2), 'program scope')
+	console.log(JSON.stringify(programTargetingCriteriaRes?.result, null, 2), 'programTargetingCriteriaRes')
+	process.exit(1)
 	if (!programTargetingCriteriaRes.success) {
 		console.error(`Failed to generate targeting criteria for program ${programIdStr}`)
 		await writeErrorRecord(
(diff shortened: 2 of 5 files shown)


src/constants/endpoints.js
```diff
@@ -19,4 +19,7 @@ module.exports = {
 	FIND_ENTITIES_BY_QUERY: process.env.FIND_ENTITIES_BY_QUERY_ENDPOINT
 		? process.env.FIND_ENTITIES_BY_QUERY_ENDPOINT
 		: 'v1/entities/find',
+	ENTITY_TYPES_FIND_BY_QUERY: process.env.ENTITY_TYPES_FIND_BY_QUERY_ENDPOINT
+		? process.env.ENTITY_TYPES_FIND_BY_QUERY_ENDPOINT
+		: 'v1/entityTypes/find',
```
adithyadinesh0412, line 24: Can we do it using the entities? Entities have api details in it for each type of entities. If it requires considerable code change we can skip that. @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -12,55 +12,74 @@ require('../../configs/events')()
...
+const missingVariables = requiredEnv.filter((key) => !process.env[key])
+
+// Throw error and exit if any required variables are missing
+if (missingVariables.length > 0) {
+	throw new Error(`Missing required environment variables: ${missingVariables.join(', ')}`)
+	process.exit(1)
 }
 
-const dbName = mongoUrl.split('/').pop()
+const { MONGODB_URL } = process.env
```
adithyadinesh0412, line 60: Are we importing only from one DB? What about Observations and survey when it comes? We will be reusing the same script and planning to update the script then? @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -70,726 +89,1098 @@ const dbName = mongoUrl.split('/').pop()
...
+		let processedCount = 0
+		let totalCount = 0
+
+		// Get total count for progress tracking
+		totalCount = await db.collection('programs').countDocuments({
+			scope: { $exists: true, $type: 'object', $ne: {} },
+			components: { $exists: true, $type: 'array', $not: { $size: 0 } },
+			tenantId: { $nin: [null, ''] },
+			orgId: { $nin: [null, ''] },
+		})
```
adithyadinesh0412: Should we first make a call to user-service and find out active tenants and org list and import only the active tenants and org? So that we don't store data of an unwanted tenant / org.  
Also , give an option to pass tenantId / org id via CLI of the script so that in future if we need to import only specific tenant / org we can do that. - If tenant and org is not passed , we can fetch the entire data from mongo. 
Use-case : 
In a saas environment only one tenant approves the use of SCP , we can import only that tenant data to SCP , rather than fetching and storing the entire DB. 
@priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -70,726 +89,1098 @@ const dbName = mongoUrl.split('/').pop()
...
-					$type: 'object',
-					$ne: {},
+		// Create efficient aggregation pipeline
+		const pipeline = [
+			{
+				$match: {
+					scope: { $exists: true, $type: 'object', $ne: {} },
+					components: { $exists: true, $type: 'array', $not: { $size: 0 } },
+					tenantId: { $nin: [null, ''] },
+					orgId: { $nin: [null, ''] },
```
adithyadinesh0412: Same comment here also to fetch tenant and org based on actual data. @priyanka-TL

adithyadinesh0412: Also , we should fetch only `isAPrivateProgram = false ` type of programs right? @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -12,55 +12,74 @@ require('../../configs/events')()
...
 }
 
-const dbName = mongoUrl.split('/').pop()
+const { MONGODB_URL } = process.env
+const dbName = MONGODB_URL.split('/').pop()
+
+// Configuration constants
+const BATCH_SIZE = 10 // Process 10 programs at a time
+const USER_CACHE_SIZE = 1000 // Cache up to 1000 users
+const CURSOR_TIMEOUT = 30 * 60 * 1000 // 30 minutes cursor timeout
```
adithyadinesh0412: Can these be a separate config file for the script? So the adopters can have control over the variables rather than making code change in the main file? @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -70,726 +89,1098 @@ const dbName = mongoUrl.split('/').pop()
...
-					csvRecords.push({
-						programId: programIdStr,
-						solutionId: '',
-						type: common.RESOURCE_TYPE_PROGRAM,
-						success: 'Program Exist',
-						resourceId: isProgramExist.resourceId,
-						rolloutId: '',
-					})
-					continue
+		const entityKeys = ['categories', 'recommended_for', 'languages']
```
adithyadinesh0412: Instead of hardcoding , can we fetch this from `entities_model_mapping` @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -1452,379 +1648,1194 @@ async function generateCertificateCriteria(
...
+ * Fetches user details for the given user IDs within a specific tenant.
+ * @async
+ * @function getUserOrgTenantDetails
+ * @param {string[]} userIds - Array of user IDs to fetch details for.
+ * @param {string} tenantCode - Code identifying the tenant.
+ * @returns {Promise<Object>} A promise that resolves to an object keyed by user ID containing user details. Returns an empty object if no users are found.
+ */
+async function getUserOrgTenantDetails(userIds, tenantCode) {
+	// Fetch user details using the user request service
+	const users = await userRequest.list('all', '', '', '', '', tenantCode, { user_ids: userIds })
```
adithyadinesh0412, line 2051: Can we cache the data in the heap / redis to reduce user-serice calls? @priyanka-TL . I don't think it will be ideal to call the user-service each time for each program / solution.

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -1452,379 +1648,1194 @@ async function generateCertificateCriteria(
...
+ * @returns {Promise<Object>} Object containing organization_code, tenant_code, user_id, and assignedTo
+ */
+async function getOrgAndTenantWithFallback(createdBy, userOrgTenantMap, programTenantId, programOrgId, orgAdminCache) {
+	// Default to program's org and tenant
+	let organization_code = programOrgId
+	let tenant_code = programTenantId
+	let user_id = createdBy
+
+	// Override with user's org and tenant if available
+	if (createdBy && userOrgTenantMap[createdBy]) {
```
adithyadinesh0412, line 1915: Same with this function also , can we introduce a look up mechanism before making api call? So that repeated call for the same data can be avoided @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
```diff
@@ -1452,379 +1648,1194 @@ async function generateCertificateCriteria(
...
+ * @param {string} tenant_code - The tenant code for database operations
+ * @returns {Promise<Object>} Object with success status and result array containing targeting criteria
  */
-async function targetedRoles(entityId, type, db) {
+async function generateTargetingCriteria(scope = {}, tenant_code) {
 	try {
-		let result = {
-			success: true,
-			targetedRoles: [],
+		if (!scope || Object.keys(scope).length === 0) {
```
adithyadinesh0412, line 2568: Same for this function , if we can have a look up mechanism before making api calls. @priyanka-TL

adithyadinesh0412 (changes requested):

src/constants/endpoints.js
priyanka-TL: If the entities are not present based on the type we need to get few details about entityType, for example spliting the role and location entities etc

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: Yes we can plan to update at a later stage,

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: We are not storing the unwanted data, we are keeping the valid data, if the user or org admin exists in that tenant, so can avoid this api call. Also here we are grouping based on tenants, hence user service call  also we will get the multiple user details in single call,

priyanka-TL: In the project tenant and org is mandatory, hence it's a one time script to migrate all data, to run this one time will be the better option I think, otherwise they have to run the script for multiple tenants and org.

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: Hence it's not possible to add this in env, i will create a config.json file is that fine

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: For validation we are adding many entityTypes in projects, hence we couldn't able to identify it, for example file type etc, hence keeping this, what I can do this I can take this from config file

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: It's actually there

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: It's there

src/scripts/elevateProject/migrateProgramsAndSolutions.js
priyanka-TL: We cannot add this on the lookup hence the entities will change based on state, and all the data stored in the memory can also lead to issues. It's hard to keep on track as per the current elevate inconsistent structure of scope.

src/scripts/elevateProject/migrateProgramsAndSolutions.js
adithyadinesh0412: We can follow the same approach , but in addition to that passing tenant / org will also be a good option is what i'm thinking. for the specific use-case I mentioned above. @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
adithyadinesh0412: Yes, we can have a common config for all the scripts , so that we can use in other scripts also if needed. @priyanka-TL

src/scripts/elevateProject/migrateProgramsAndSolutions.js
adithyadinesh0412: This will be a one time script which will be executed, so fetching the entities for Karnataka for project 1 will be same for project 2. 
We can have a heap storage or use redis 
Keeping in mind the scale of resources , hitting external apis might overwhelm the other services, just a hashmap or simple map of for example 
` entityLookUp =  { 
karnataka-state : _id ,
bangalore-district : _id,
student-professional-sub-role : _id
}
`
if(entityLookUp.?[${value}-${entityType}]) {
 return entityLookUp.?[${value}-${entityType}]
} else {
 
  // fetch from api
}

@priyanka-TL

PR #531, opened 2025-11-05
Add TENANT_ADMIN_ROLE with permissions and async tenant validation

adithyadinesh0412:
## Summary by CodeRabbit

* **New Features**
  * Added a tenant-admin role and a new internal tenant endpoint.
  * Migration added to provision tenant-admin permissions.

* **Enhancements**
  * Async internal tenant validation to improve tenant/org resolution and clearer error responses.
  * Added localization entries for tenant/org not found.

* **Bug Fixes**
  * Migration now avoids creating duplicate modules during provisioning.

### DIFF
--- a/src/generics/utils.js
+++ b/src/generics/utils.js
@@ -854,50 +854,55 @@ const fetchInternalTenantDetails = async function (tenantCode) {
  * @throws {Error} - If tenant or organization validation fails
  */
 async function _extractTenantAndOrgCodes(req) {
-	let tenantCode = req.decodedToken.tenant_code
-	let organizationCode = req.decodedToken.organization_code
-	// if tenant admin role, validate if organization exists within tenant
-	// return the org code from header else from token
-	// tenant passed in the header will be ignored
-	if (validateRoleAccess(req.decodedToken.roles, common.TENANT_ADMIN_ROLE)) {
-		organizationCode =
-			req.headers?.[process.env.ORG_ID_HEADER_NAME.toLocaleLowerCase()] || req.decodedToken.organization_code
-		const fetchTenantData = await fetchInternalTenantDetails(tenantCode)
-		if (fetchTenantData.success) {
-			const orgWithInTenant = fetchTenantData.data.organizations
-				? fetchTenantData.data.organizations.find((org) => org.code === organizationCode)
-				: null
-			if (!orgWithInTenant) {
+	try {
+		let tenantCode = req.decodedToken.tenant_code
+		let organizationCode = req.decodedToken.organization_code
+		// if tenant admin role, validate if organization exists within tenant
+		// return the org code from header else from token
+		// tenant passed in the header will be ignored
+		if (validateRoleAccess(req.decodedToken.roles, common.TENANT_ADMIN_ROLE)) {
+			organizationCode =
+				req.headers?.[process.env.ORG_ID_HEADER_NAME.toLocaleLowerCase()] || req.decodedToken.organization_code
+			const fetchTenantData = await fetchInternalTenantDetails(tenantCode)
+			if (fetchTenantData.success) {
+				const orgWithInTenant =
+					fetchTenantData?.data?.organizations && Array.isArray(fetchTenantData.data.organizations)
+						? fetchTenantData.data.organizations.find((org) => org.code === organizationCode)
+						: null
+				if (!orgWithInTenant) {
+					return {
+						error: 'ORGANIZATION_NOT_FOUND_IN_TENANT',
+					}
+				}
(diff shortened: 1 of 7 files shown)


src/generics/utils.js
```diff
@@ -802,12 +803,83 @@ function validateTenantAndOrganizationInHeader(req) {
...
-
+	// if tenant admin role, validate if organization exists within tenant
+	// return the org code from header else from token
+	// tenant passed in the header will be ignored
+	if (validateRoleAccess(req.decodedToken.roles, common.TENANT_ADMIN_ROLE)) {
+		organizationCode =
+			req.headers?.[process.env.ORG_ID_HEADER_NAME.toLocaleLowerCase()] || req.decodedToken.organization_code
+		const fetchTenantData = await fetchInternalTenantDetails(tenantCode)
+		if (fetchTenantData.success) {
+			const orgWithInTenant = fetchTenantData.data.organizations
```
priyanka-TL: @adithyadinesh0412 Please validate data and organizations are exist

src/generics/utils.js
```diff
@@ -802,12 +803,83 @@ function validateTenantAndOrganizationInHeader(req) {
...
+ * @param {Object} req.headers - Request headers
+ * @returns {Promise<Object>} Returns object with structure:
+ *   {
+ *     tenantCode: String,        // Validated tenant code
+ *     organizationCode: String,  // Validated organization code
+ *     error?: String            // Error message if validation fails
+ *   }
+ * @throws {Error} - If tenant or organization validation fails
+ */
+async function _extractTenantAndOrgCodes(req) {
```
priyanka-TL, line 856: @adithyadinesh0412 Can we use try catch here

src/generics/utils.js
```diff
@@ -802,12 +803,83 @@ function validateTenantAndOrganizationInHeader(req) {
...
+const fetchInternalTenantDetails = async function (tenantCode) {
+	let result = {
+		success: false,
+		data: {},
+	}
+	try {
+		const userBaseUrl = buildUrl(process.env.INTERFACE_SERVICE_HOST, process.env.USER_SERVICE_BASE_URL)
+		const tenantReadUrl = buildUrl(userBaseUrl, endpoints.FETCH_TENANT_INTERNAL, {}, tenantCode)
+		const tenantDetails = await requests.get(tenantReadUrl, '', true)
+		if (tenantDetails && tenantDetails?.success) {
```
priyanka-TL, line 827: @adithyadinesh0412 if the tenantDetails is empty or success is false can we throw error ?

src/generics/utils.js
```diff
@@ -802,12 +803,83 @@ function validateTenantAndOrganizationInHeader(req) {
...
+			}
+		} else {
+			return {
+				error: 'TENANT_NOT_FOUND',
+			}
+		}
+	}
+	// if super admin role, fetch tenant and organization from headers
+	// admin will override the tenant_admin privilages if the user have both rights
 	if (validateRoleAccess(req.decodedToken.roles, common.ADMIN_ROLE)) {
```
priyanka-TL: @adithyadinesh0412 instead of taking the admin role from common can we take from env ? what if we configured a different role title for all roles ?

priyanka-TL (changes requested):
Reviewed on 1 PM

src/generics/utils.js
adithyadinesh0412: This function is written as a generic fetch function. 
If the tenant is not found , we are returning `success = false` and `data = []`
which is checked in the invoking function and return relevant error based on the use-case. 

@priyanka-TL 
If you see in this case , below code ensures to throw relevant error in function  `_extractTenantAndOrgCodes(req)`

`const fetchTenantData = await fetchInternalTenantDetails(tenantCode)
			if (fetchTenantData.success) {
				const orgWithInTenant = fetchTenantData?.data?.organizations && Array.isArray(fetchTenantData.data.organizations)
					? fetchTenantData.data.organizations.find((org) => org.code === organizationCode)
					: null
				if (!orgWithInTenant) {
					return {
						error: 'ORGANIZATION_NOT_FOUND_IN_TENANT',
					}
				}
			} else {
				return {
					error: 'TENANT_NOT_FOUND',
				}
			}`

src/generics/utils.js
adithyadinesh0412: Every other service , we are following the same flow. I don't think we need to add an additional `.env` constraint for this. 
Also , configuring admin role shouldn't be allowed according to my opinion as admin privileges are limited. @priyanka-TL

priyanka-TL (approved):

